PR #2, opened 2015-09-10
Fix xattr support on linux

dmcgowan:
Remove support for xattr on symbolic links due to lack of syscall support in golang system package.

Correct xattr support on symlinks requires support for the `llistxattr` syscall. Since this is currently missing is makes sense to remove support for xattr on symlinks, as already questioned by the comment above it.


stevvooe:
Not exactly "fix" but will do it for now...

PR #3, opened 2015-09-11
Add context interface

dmcgowan:
Add context interface which is used to build manifests

### DIFF
--- a/context.go
+++ b/context.go
@@ -16,10 +16,17 @@ var (
 	ErrNotFound = fmt.Errorf("not found")
 )
 
-// Context represents a file system context for accessing resources.
+type Context interface {
+	Verify(Resource) error
+	Resource(string, os.FileInfo) (Resource, error)
+	Sanitize(string) (string, error)
+	Walk(filepath.WalkFunc) error
+}
+
+// context represents a file system context for accessing resources.
 // Generally, all path qualified access and system considerations should land
 // here.
-type Context struct {
+type context struct {
 	root string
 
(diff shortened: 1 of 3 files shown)


context.go
```diff
@@ -30,7 +37,7 @@ type Context struct {
 }
 
 // NewContext returns a Context associated with root.
-func NewContext(root string) (*Context, error) {
+func NewPathContext(root string) (Context, error) {
```
stevvooe: I don't like this change. This will always be `NewContext`. I'll just merge and fix, but your code will break.

manifest.go
```diff
@@ -1,37 +1,30 @@
...
-			return nil
+		if err != nil {
+			return err
 		}
 
 		sanitized, err := ctx.Sanitize(p)
 		if err != nil {
 			return err
 		}
+		if sanitized == "." {
```
stevvooe: The root should never just be ".". It should always be invoked as a full path. This doesn't work correctly for all cases.

context.go
dmcgowan: Yeah, makes sense, I will keep as `NewContext`

manifest.go
dmcgowan: Then I may have gotten the logic wrong, in the test I ran I saw it show up as '.'.  Let me explore what those other cases are

stevvooe: I think the issue is that this interface does not properly contain the context. `Walk` here exposes paths outside the context. This will become better with better isolation here.

PR #6, opened 2015-09-17
Add diff util to differ manifests

dmcgowan:
Provides a mechanism for using continuity to diff manifests

* This could go in a separate package altogether since it is a utility using continuity rather than a utility used by continuity.

### DIFF
--- a/diffutil/diff.go
+++ b/diffutil/diff.go
@@ -1,0 +1,159 @@
+package diffutil
+
+import (
+	"github.com/Sirupsen/logrus"
+	"github.com/stevvooe/continuity"
+)
+
+type ResourceUpdate struct {
+	Original continuity.Resource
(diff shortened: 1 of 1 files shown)


stevvooe:
I like this concept. I'm wondering if this should be a first-class feature of the manifest or not. Should we have resource comparison functions as part of the exported API? I think so!

In reality, I do have some local work that goes in this direction to implement `Apply`. Since `Apply` is effectively, diff with the current state then apply the delta, it can work very similarly.

Thoughts?

stevvooe:
Should we put this on the manifest type?

dmcgowan:
I am going to pull this in for unit testing now then we can expose later if we find it more widely useful. Going to now expose as diffing Resource Lists.

PR #9, opened 2016-01-07
Add full implementation of xattr for linux and darwin

dmcgowan:
Include support for no follow symbolic link on both systems

Cleans up continuity interface to move internal syscall functions to `sysx` package.


stevvooe:
LGTM

I'll give this a whirl on darwin when I have a chance.

PR #10, opened 2016-01-08
Add implementation of context apply

dmcgowan:
Apply sets file properties for a resource to a file within a context. `ApplyManifest` can be used to apply all the resources in a manifest.

Note: symlink behavior requires more research and testing.


driver_unix.go
```diff
@@ -4,11 +4,27 @@ package continuity
...
 
 	"github.com/stevvooe/continuity/sysx"
 )
 
+const (
+	atNoFollow = 0x100
+)
+
+func (d *driver) Unlink(path string) error {
+	return syscall.Unlink(path)
```
stevvooe: Why not `os.Remove`?

driver_unix.go
```diff
@@ -4,11 +4,27 @@ package continuity
...
+}
+
+// Lchmod changes the mode of an file not following symlinks. The name
+// must be an absolute path to a file.
+func (d *driver) Lchmod(name string, mode os.FileMode) error {
+	return syscall.Fchmodat(0, name, uint32(mode), atNoFollow)
```
stevvooe: This is not a correct use of `Fchmodat`, if I understand correctly. This will do the operation relative to the file descriptor in the first argument.

context.go
```diff
@@ -339,3 +463,21 @@ func (c *context) resolveXAttrs(fp string, fi os.FileInfo, base *resource) (map[
...
+// from the given linkname.
+func RelativeSymlinkPath(root, linkname, target string) (string, error) {
+	return filepath.Rel(filepath.Join(root, linkname), target)
+}
+
+// ChrootSymlinkPath uses the given absolute target intended for a
+// chroot environment using the given root.
+func ChrootSymlinkPath(root, linkname, target string) (string, error) {
+	return target, nil
+}
```
stevvooe: Do we have a namespace version of these?

dmcgowan: Currently only `AbsoluteSymlinkPath` is used, the others are added to demonstrate how it may be used in the future. It would involve having the context configurable. For a namespace version, not sure if additional properties would be needed, as we previously discussed the namespace case would ideally use the `AbsoluteSymlinkPath` and the re-map would not be the responsibility of continuity.

driver_unix.go
dmcgowan: The intent was to only unlink and remove a file rather than remove a directory. They two remove operations could be joined like they are in the `os` package.

driver_unix.go
dmcgowan: From linux kernel description
>  If pathname is absolute, then dirfd is ignored

From BSD description (could not find directly in darwin, maybe darwin has something akin to lchmod)
> The fchmodat() is equivalent to either chmod() or lchmod()	depending on the flag except in the case where path specifies a relative path.

Fchmodat should operate the same as `chmod` when the given fd is 0. There is a comment about how this function must take in an absolute path. I could explicitly check and error here if it is not. On linux there is `lchmod` so this is the correct system call to use.

dmcgowan:
Updated to address comments

- Now single higher level `Remove` function in driver
- Added `sysx` implementation for `Fchmodat` on darwin
- `Lchmod` will now ensure paths are absolute before calling `Fchmodat`

stevvooe:
LGTM

PR #11, opened 2016-01-09
Import sha256 to avoid panic from digest package

dmcgowan:
Digest package used to panic with nil pointer exception but now has an explicit panic. This explicit import will allow continuity with the latest digest package version.


stevvooe:
LGTM

PR #14, opened 2016-01-12
Add unit testing

dmcgowan:
Update the unit testing to check the resources from the manifest.
Add support for Linux device info (needed for testing but support for device numbers may be dropped altogether in the future).
Fix bug with absolute symlinks getting wrong target in created Resource.

### DIFF
--- a/manifest_test.go
+++ b/manifest_test.go
@@ -2,12 +2,16 @@ package continuity
 
 import (
 	_ "crypto/sha256"
+	"errors"
 	"fmt"
 	"io/ioutil"
 	"math/rand"
 	"os"
 	"os/exec"
 	"path/filepath"
+	"sort"
+	"strconv"
+	"syscall"
 	"testing"
 
 	"github.com/docker/distribution/digest"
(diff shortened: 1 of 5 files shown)


manifest_test.go
```diff
@@ -259,3 +302,117 @@ func randomBytes(p []byte) {
...
+	paths := []string{}
+	for _, r := range resources {
+		absPath := r.path
+		if !filepath.IsAbs(absPath) {
+			absPath = "/" + absPath
+		}
+		uidStr := strconv.Itoa(r.uid)
+		gidStr := strconv.Itoa(r.uid)
+		switch r.kind {
+		case rfile:
```
stevvooe: So, it would be nice to get rid of these test types. I actually made them before the advent of the default package schema. I think this use case could be filled in with a declarative resource struct, coupled with a method that will select the correct wrapping type:

```go
r, err := Declare(Resource{
    Kind: RegularFile,
    Path: "/foo",
})
```

And for inline declarations for making tests:

```go
[]Resource{
    MustDeclare(Resource{
        Kind: RegularFile,
        Path: "/foo",
    }),
}
```

stevvooe:
LGTM

My comment is really to inform refactoring of test code. Accepting this as is.

manifest_test.go
dmcgowan: Created issue #15 to track this

PR #19, opened 2016-01-20
Readonly fuse fs

dmcgowan:
Adds a read only fuse filesystem created from the manifest and getting content from a directory.


stevvooe:
LGTM

PR #20, opened 2016-01-21
Group user name support

dmcgowan:
This PR re-adds user and group name support. This PR needs more testing but should be discussed to ensure we get the preference right as well as intended encoding behavior.

This PR prefers to use user and group names when provided over the given UID/GID. The most lacking part of this change I believe is that additional errors looking up UID/GID are not handled or logged, and cases when UID/GID mismatch could be detected nothing is done to mitigate or inform. I am not sure that there is anything reasonable we can do here.


group/group.go
```diff
@@ -0,0 +1,35 @@
+// Copyright 2011 The Go Authors. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
```
aaronlehmann: Could you please add a comment on the lineage of these files and what modifications were made?

resource.go
```diff
@@ -212,13 +235,35 @@ func (r *resource) Mode() os.FileMode {
 }
 
 func (r *resource) UID() string {
+	if r.user != "" {
+		u, err := user.Lookup(r.user)
```
aaronlehmann: I think it's a good idea for Continuity to cache the results of these user and group lookups. The system library calls could be quite expensive when multipled by hundreds of thousands of files; I don't know.

aaronlehmann:
I'm +1 on this concept. I think it's better to track users and groups by name when possible; otherwise a continuity manifest is tied to the system that created it.

group/group.go
dmcgowan: Added "Derived from" comment

dmcgowan:
Added a group/user cache. This is assuming that it is ok that user changes done after continuity is invoked are not reflected. In the case where a user did not exist but was added while the application was running, continuity would use the specified uid.

aaronlehmann:
I think it's okay to ignore changes to the user/group database after continuity is invoked. I considered this when I made the recommendation, and thought that user/group changes made on the system during an apply operation would lead to inconsistent results, so it's probably best to ignore these anyway. If continuity ends up being used in a long-running way that makes this cache persistent, we may want to revisit this.

aaronlehmann:
After some discussion with @tonistiigi, I've come to think that it actually makes more sense to use UIDs/GIDs. The reason is that it doesn't matter these UIDs/GIDs map to on the host. The filesystem being captured may have its own /etc/passwd that ascribes special significance to particular UIDs, and we don't want to interfere with that by remapping IDs.

Sorry for requesting something that I ultimately changed my mind on.

stevvooe:
@aaronlehmann Generally, I favor the UID/GID model over the uname/gname model. In practice, I am not sure what will actually work.

Given your statement, do we close this?

aaronlehmann:
I don't mind this feature being added, but sadly I no longer have a use case for it.

stevvooe:
After a long consideration, we should revisit this. Seems like this patch still applies. Any reservations?

stevvooe:
@dmcgowan After having some experience elsewhere, I think we can do GID/UID and group/supplementary support.

We should follow the `<user>[:<group>` syntax for user + primary group, then include a supplementary groups field. From there, we can define which to use. There is the problem that we can't really assume the existence of an `/etc/passwd` file, so users will have to be careful when porting from system to system.

PR #22, opened 2016-01-29
Skip chmod on symlinks

dmcgowan:
No follow for chmod on symlinks is currently not implemented in the linux kernel. Skip to prevent unintended chmod on targets.


stevvooe:
What a bazaar omission.

LGTM

PR #23, opened 2016-01-29
Fix fuse hardlinks

dmcgowan:
Adds multiple paths to the filesystem for each hardlink, maintaining the original inode number.


stevvooe:
LGTM

PR #24, opened 2016-02-04
apply: correct handling of symlinks

stevvooe:
Previously, tricky things were attempted with symlinks to make filesystems more
portable. Without operating system support, this effort was wasted and caused
bugs. Code attempting to fixup symlinks has been elided in favor of simply
transporting the link targets as is.

A small bug in lchown was fixed, as well.

Signed-off-by: Stephen J Day <stephen.day@docker.com>

cc @dmcgowan @dmcgowan

### DIFF
--- a/context.go
+++ b/context.go
@@ -37,9 +37,8 @@ type SymlinkPath func(root, linkname, target string) (string, error)
 // Generally, all path qualified access and system considerations should land
 // here.
 type context struct {
-	driver  Driver
-	root    string
-	symPath SymlinkPath
+	driver Driver
+	root   string
 }
 
 // NewContext returns a Context associated with root. The default driver will
@@ -70,9 +69,8 @@ func NewContext(root string) (Context, error) {
 	}
 
 	return &context{
-		root:    root,
-		driver:  driver,
-		symPath: AbsoluteSymlinkPath,
+		root:   root,
+		driver: driver,
 	}, nil
 }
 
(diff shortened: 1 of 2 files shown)


context.go
```diff
@@ -321,42 +306,40 @@ func (c *context) Apply(resource Resource) error {
...
 		}
+
+		// NOTE(stevvooe): Chmod on symlink is not supported on linux. We
+		// may want to maintain support for other platforms that have it.
 		chmod = false
 	}
 
 	// Update filemode if file was not created
-	if chmod && exists {
+	if chmod {
```
aaronlehmann: This change causes a redundant syscall when creating a new directory that didn't exist before. The `Mkdir` call includes the desired mode, and then `Lchmod` is called here. Previously, `exists` would be set to false, so `Lchmod` would not be called.

I'm not sure this is a big deal.

aaronlehmann:
LGTM, though we may want to fix the redundant syscall I pointed out.

I also would like to fix `Apply` to replace a regular file with a symlink when necessary. I will do this in a separate PR.

context.go
stevvooe: This code needs some general syscall optimization but I wanted to avoid the refactoring to get there.

stevvooe:
> I also would like to fix Apply to replace a regular file with a symlink when necessary. I will do this in a separate PR.

This PR does that but it doesn't handle an existing directory. It uses the existence of `target` to decide whether or not to remove the file. This worked on my local system.

We really need to setup verification tests.

aaronlehmann:
> This PR does that but it doesn't handle an existing directory. It uses the existence of target to decide whether or not to remove the file. This worked on my local system.

Ah yes, noticed that after posting the comment. I'm working on a few enhancements to `Apply` and `Verify` right now that deal with similar issues.

If it's okay with you, I'll merge this PR first.

Totally agree about tests.

stevvooe:
Merging.

PR #25, opened 2016-02-04
On apply, verify that regular files are regular files, and directories are directories

aaronlehmann:
Summary says it all.


stevvooe:
LGTM

PR #34, opened 2016-02-09
Add support for applying and verifying devices and named pipes

aaronlehmann:
Since devices and named pipes can support hard links, this involves
changing the Merge function to support types other than RegularFile.

This also fixes Verify to properly check for hard links. Part of Verify
is split into verifyMetadata. This portion is called for each link, but
digests are only verified on one instance of the inode.


stevvooe:
LGTM

PR #35, opened 2016-06-09
Add data interfaces

dmcgowan:
Add interfaces for a context to be able to use a blob store for storing and fetching content. The continuity interface does not need to expose the concept of the blob store since it is only concerned about bytes and hashes and nothing related to their storage.

This updates the Apply logic to be able to set the content of a file, creating files that do not exist. Currently the apply is still lacking the functionality to completely sync a directory, which could include removing files. It may make sense to split off the currently Apply (just metadata) and Sync (changes content).

### DIFF
--- a/context.go
+++ b/context.go
@@ -33,26 +33,45 @@ type Context interface {
 // not under the given root.
 type SymlinkPath func(root, linkname, target string) (string, error)
 
+type ContextOptions struct {
+	Digester Digester
+	Driver   Driver
+}
+
 // context represents a file system context for accessing resources.
 // Generally, all path qualified access and system considerations should land
 // here.
 type context struct {
-	driver Driver
-	root   string
+	driver   Driver
+	root     string
+	digester Digester
 }
 
 // NewContext returns a Context associated with root. The default driver will
 // be used, as returned by NewDriver.
 func NewContext(root string) (Context, error) {
+	return NewContextWithOptions(root, ContextOptions{})
+}
+
+// NewContextWithOptions returns a Context associate with the root.
+func NewContextWithOptions(root string, options ContextOptions) (Context, error) {
 	// normalize to absolute path
 	root, err := filepath.Abs(filepath.Clean(root))
 	if err != nil {
 		return nil, err
 	}
 
-	driver, err := NewSystemDriver()
-	if err != nil {
-		return nil, err
+	driver := options.Driver
+	if driver == nil {
+		driver, err = NewSystemDriver()
+		if err != nil {
+			return nil, err
+		}
+	}
+
+	digester := options.Digester
+	if digester == nil {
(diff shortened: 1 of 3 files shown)


digests.go
```diff
@@ -8,18 +8,24 @@ import (
...
+}
 
-	f, err := d.Open(p)
-	if err != nil {
-		return "", err
-	}
-	defer f.Close()
+// ContentProvider produces a read stream for a given digest
+type ContentProvider interface {
+	Reader(digest.Digest) (io.Reader, error)
```
stevvooe: Nit: should this be an `io.ReadCloser`?

context.go
```diff
@@ -337,6 +361,39 @@ func (c *context) Verify(resource Resource) error {
...
+		r, err = c.provider.Reader(dgst)
+		if err == nil {
+			break
+		}
+	}
+	if err != nil {
+		return fmt.Errorf("file content could not be provided: %v", err)
+	}
+
+	f, err := os.OpenFile(fp, flag, rf.Mode())
```
stevvooe: Needs something like this: https://github.com/docker/swarmkit/blob/f0983eb859019cf5dcfef9ef3baa2facb8a1c4f5/ioutils/ioutils.go#L13.

Basically, make a _sibling_ tempfile, write all data, flush, then move.

context.go
```diff
@@ -337,6 +361,39 @@ func (c *context) Verify(resource Resource) error {
...
+	}
+	var (
+		r   io.Reader
+		err error
+	)
+	for _, dgst := range rf.Digests() {
+		r, err = c.provider.Reader(dgst)
+		if err == nil {
+			break
+		}
```
stevvooe: I think we end up danging without readcloser here.

stevvooe:
All in all, this looks good. There may be some tricks around writing files we may need to handle. I'm still not sure exactly what the appropriate apply model should be, but this is a good start.

dmcgowan:
Updated to do the atomic writes and enforce closing.

As for `Apply`, I was tempted to split out the existing apply logic from this newer version which does checkout. The concern I had is around file content, currently the `Apply` function does not check digests making the function not that practical in real usage scenarios. As soon as digests are checked, then it makes sense to set the content if there is a content provider. Likewise if there is no content provider and the digests match, `Apply` should fail as the filesystem does not represent the desired state. Along these lines, we probably need to update the logic to remove files which are not expected. I think we want to guarantee at the end that if you were to generate a manifest over an applied directory you would get the same manifest which was applied without extra files.

context.go
stevvooe: @dmcgowan Do we need to close the read closer?

stevvooe: Ah, I see, this breaks when the file is open.

stevvooe:
LGTM

PR #38, opened 2016-09-22
Update fuse interface to use io.ReaderCloser

dmcgowan:
Allows interface to be implemented from outside the continuityfs package.


stevvooe:
LGTM

PR #39, opened 2016-09-22
Add manifest merge logic for applying fs diffs

dmcgowan:
Allow creating manifest directly from an overlay or aufs diff directory and merging on top of another manifest.

This is an initial implementation, further optimization and unit testing should be considered for wider use. The most useful application could be turning Docker image layers directly into a continuity manifest for an image (after converting each tar into manifest).

### DIFF
--- a/merge.go
+++ b/merge.go
@@ -1,0 +1,96 @@
+package continuity
+
+import (
+	"sort"
+	"strings"
+)
+
+// MergeManifests merges a manifest onto another manifest. Only
+// the diff manifest should contain whiteout information.
+func MergeManifests(manifest, diff *Manifest) *Manifest {
+	r1 := manifest.Resources
+	sort.Sort(ByPath(r1))
+	r2 := diff.Resources
+	sort.Sort(ByPath(r2))
+	return mergeResources(r1, r2)
+}
+
+// mergeResources merges two ordered set of resources into manifest
+// using the provided whiteout and comparison function.
+// TODO(dmcgowan): Handle handlinks
+func mergeResources(r1, r2 []Resource) *Manifest {
+	result := make([]Resource, 0, len(r1))
+
+	i1 := 0
+	i2 := 0
+
+	for i1 < len(r1) && i2 < len(r2) {
+		p1 := r1[i1].Path()
+		p2 := r2[i2].Path()
+
+		switch {
+		case p1 < p2:
+			result = append(result, r1[i1])
+			i1++
+		case p1 == p2:
+			// p1 will be replaced by p2
+			i1++
+			fallthrough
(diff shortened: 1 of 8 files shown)


merge.go
```diff
@@ -0,0 +1,226 @@
+package continuity
+
+import (
+	"log"
+	"path/filepath"
+	"sort"
+	"strings"
+)
+
+// MergeOverlay merges a manifest created from an overlay diff
```
stevvooe: How are details of the overlay implementation making it up to the manifest level?

stevvooe:
@dmcgowan Should continuity define its own diff format rather than letting aufs/overlay implementations escape?

dmcgowan:
I think we should choose one method and use it. Docker "choose" using the aufs whiteout method. Personally I think the overlay whiteout method is far superior, it is much easier to process since whiteouts can be immediately actionable rather than waiting to come across a whiteout file. In overlay whiteouts are simply devices with major and minor numbers of 0, and the opaqueness is handled as an xattr. If we want to convert, we could easily change the name of the xattr to be continuity specific.

I wanted to get a proof a concept working using Docker diffs so I added the AUFS implementation. You can see the overlay implementation is much more straightforward. Instead what I will do is drop the AUFS implementation and do whiteout conversion during the tar conversion (this is what Docker does for Overlay2).

I don't have a problem with sticking with overlay's whiteout method, the documentation is very short and easy to understand. I don't need to mention Overlay in the function name though if we ditch AUFS specific logic.

```
whiteouts and opaque directories
--------------------------------

In order to support rm and rmdir without changing the lower
filesystem, an overlay filesystem needs to record in the upper filesystem
that files have been removed.  This is done using whiteouts and opaque
directories (non-directories are always opaque).

A whiteout is created as a character device with 0/0 device number.
When a whiteout is found in the upper level of a merged directory, any
matching name in the lower level is ignored, and the whiteout itself
is also hidden.

A directory is made opaque by setting the xattr "trusted.overlay.opaque"
to "y".  Where the upper filesystem contains an opaque directory, any
directory in the lower filesystem with the same name is ignored.
```

stevvooe:
@dmcgowan For tar files, I agree that either encoding is fine. However, for continuity, we control the format. Let's define a diff manifest type that calls these out explicitly, along with careful ordering constraints.

dmcgowan:
Pushed an update I worked on before. Adds a whiteout type which allows creation of these diff manifests without relying on an external format. The merge logic now only applies a diff manifest to another manifest to produce the merged result. Support for AUFS and Overlay diffs will be done by implementing the defined interface for context to produce the correct resources. Ordering is already handled through the manifest ordering.

PR #41, opened 2016-10-11
Fix bugs in applying files

dmcgowan:
When no digest is given simply create and close a new file. Previously creating a file with no digest caused a panic since the reader is not set.

For setting up symlinks, the original link should only be attempted to be removed if there was file info.


stevvooe:
@dmcgowan How can a regular file have no digests? This seems like a corrupt manifest.

dmcgowan:
I think that is fair, there is a better solution which involves special casing empty files with the empty digest. The tar-split logic skips processing of empty files in what I was working on so I was getting an error with these files.

PR #45, opened 2017-02-01
Add copy file range to sysx

dmcgowan:
Allows copying a file in kernel and allows the kernel to use reflink copying if supported.


dmcgowan:
Tested copying file on btrfs and ext4. The copy on btrfs is instantaneous because it is able to use a reflink copy. This only calls `sysx.CopyFileRange`.

```
$ time ./copyfilerange btrfs/source1 btrfs/copy1
./copyfilerange btrfs/source1 btrfs/copy1  0.00s user 0.00s system 95% cpu 0.002 total
$ ls -l btrfs/      
total 4194296
-rw-r--r--. 1 derek derek 2147479552 Jan 31 17:29 copy1
-rw-r--r--. 1 derek derek 2147479552 Jan 31 17:19 source1


$ time ./copyfilerange ext4/source1 ext4/copy1  
./copyfilerange ext4/source1 ext4/copy1  0.00s user 2.48s system 3% cpu 1:22.35 total
$ ls -l ext4
total 4194304
-rw-r--r--. 1 derek derek 2147479552 Jan 31 17:31 copy1
-rw-r--r--. 1 derek derek 2147479552 Jan 31 17:16 source1
```

dmcgowan:
It also works with the new xfs shared extent features in 4.9. Formatted xfs with `mkfs.xfs -m reflink=1`.

```
$ time ./copyfilerange xfs/source1 xfs/copy1                   
./copyfilerange xfs/source1 xfs/copy1  0.00s user 0.00s system 20% cpu 0.024 total
$ ls -l xfs
total 4194296
-rw-r--r--. 1 derek derek 2147479552 Jan 31 21:49 copy1
-rw-r--r--. 1 derek derek 2147479552 Jan 31 21:48 source1
```

dmcgowan:
Update to just directly expose generate function without helpers. Also added `arm64` implementation. I could not find a syscall number I was certain about for `arm`.

stevvooe:
LGTM

I am not sure how stable we can keep this. The `len` type probably isn't quite right, but we can play around with it later.

Issue #12, opened 2016-01-09
Use go generate in sysx


dmcgowan:
Update the generation process in sysx to take advantage of go generate

PR #51, opened 2017-02-20
sysx: add copy_linux_arm.go and sysnum_linux_arm.go

AkihiroSuda:
For https://github.com/docker/containerd/issues/528

(TODO: use go generate #12)


stevvooe:
LGTM

Do we have a way to test this?

AkihiroSuda:
We could use QEMU, but I think checking with eyes is ok for this

stevvooe:
LGTM

@dmcgowan PTAL

dmcgowan:
This needs `SYS_COPY_FILE_RANGE` defined in `sysnum_linux_arm.go`. I didn't generate this one because I couldn't find a confident source after a few minutes of looking. If you find it, please link to where you find it. Otherwise I might fire up my raspberry pi and have a look.

AkihiroSuda:
`/usr/include/arm-linux-gnueabihf/asm/unistd.h` says:

```c
#define __NR_OABI_SYSCALL_BASE	0x900000

#if defined(__thumb__) || defined(__ARM_EABI__)
#define __NR_SYSCALL_BASE	0
#else
#define __NR_SYSCALL_BASE	__NR_OABI_SYSCALL_BASE
#endif
...
#define __NR_copy_file_range	(__NR_SYSCALL_BASE+391)
```

https://github.com/golang/go/wiki/GoArm#supported-operating-systems says Go requires EABI

> You must run an EABI kernel


So, overall, it is 391.
I'll update PR in short

AkihiroSuda:
updated PR

AkihiroSuda:
@dmcgowan PTAL?

dmcgowan:
LGTM

PR #57, opened 2017-04-03
Serialize XAttrs/ADS as ordered in protos

sul3n3t:
XAttrs/ADS were previously stored as a map in proto form. There is no defined ordering of maps, which can cause inconsistent binary representation of resources with multiple attrs/ads. This commit changes the types to ordered lists. The representation should be compatible.


sul3n3t:
@stevvooe PTAL

resource.go
```diff
@@ -482,7 +482,17 @@ func toProto(resource Resource) *pb.Resource {
 	}
 
 	if xattrer, ok := resource.(XAttrer); ok {
-		b.Xattr = xattrer.XAttrs()
+		// Sorts the XAttrs by name for consistent ordering.
+		keys := []string{}
+		xattrs := xattrer.XAttrs()
```
stevvooe: Maybe, we change the XAtterer interface to return and guarantee the order?

proto/manifest.proto
```diff
@@ -9,6 +9,8 @@ message Manifest {
 }
 
 message Resource {
+    reserved 12;
```
stevvooe: I don't _think_ we need to worry about breaking. The usage here is thin on xattrs. However, the types are actually equivalent.

proto/manifest.proto
```diff
@@ -59,19 +61,19 @@ message Resource {
...
-    // TODO(stevvooe): The use of maps here may be problematic for
-    // deterministic generation. Check out this comment:
-    // https://developers.google.com/protocol-buffers/docs/proto3#backwards-compatibility
-    // Fortunately, the Go implementation correctly sorts the map keys to
-    // ensure deterministic generation, but this is not guaranteed for all
-    // implementations. If this is problem, we should generate that schema and
-    // sort by key. We can do this at any time and retain backwards
-    // compatibility.
+    // Ads stores one or more alternate data streams for the target resource.
+    map<string, bytes> ads = 13;
```
stevvooe: Do you mind doing this for ADS, as well?

sul3n3t: Sure

stevvooe:
LGTM

These should be backwards compatible with the map type. If anyone has problems, please let us know.

PR #58, opened 2017-04-06
CAS representation for alternative data stream

### LINKED ISSUE #47: [Suggestion] CAS representation for alternative data stream
```pb
message Resource {
  ..
  // Ads stores one or more alternate data streams for the target resource.
  map<string, bytes> ads = 13;
}
```

In current design, the representation of ADS is likely to become so huge, because it contains the full data of ADS.

So WDYT about adding CAS representation for ADS?
i.e.
```pb
message Resource {
  ..
  // Ads stores one or more alternate data streams for the target resource.
  map<string, bytes> ads = 13;
  // AdsDigests stores digests of one or more alternate data streams for the target resource.
  // How to access the actual data using the digest is not defined.
  // Intersection of Ads and AdsDigests SHOULD be an empty set.
  map<string, string> ads_digests = 14;
}
```

AkihiroSuda:
Fix #47 


stevvooe:
LGTM

Does this need code to handle extraction and application?

dmcgowan:
LGTM

AkihiroSuda:
@stevvooe 
Yes, but I think we can defer implementation until someone needs this

stevvooe:
@AkihiroSuda Agreed!

PR #60, opened 2017-04-06
Fix apply on non-existent symlink

dmcgowan:
When applying a symlink, skip removal if the file does not exist. This fixes a bug where apply was failing when the file did not already exist.


stevvooe:
LGTM

PR #61, opened 2017-04-07
[WIP] Add copy to method to content provider

dmcgowan:
Allows content provider to optimize creation of files directly from the blobstore. This could be used to call copy file range on file systems which support it.

### DIFF
--- a/copy_unsupported.go
+++ b/copy_unsupported.go
@@ -1,0 +1,12 @@
+// +build !linux
+
+package continuity
+
+import (
+	"io"
+	"os"
+)
+
+func fastCopy(dst *os.File, r io.Reader, size int64) (int64, error) {
+	return 0, nil
+}
--- a/digests.go
+++ b/digests.go
@@ -1,6 +1,7 @@
 package continuity
 
 import (
+	ctxu "context"
 	"fmt"
 	"io"
 	"sort"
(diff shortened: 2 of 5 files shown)


digests.go
```diff
@@ -13,23 +13,23 @@ type Digester interface {
...
-// ContentProvider produces a read stream for a given digest
+// ContentProvider provides content for a digest.
 type ContentProvider interface {
+	// Reader returns a stream of the content for
+	// the provided digest.
 	Reader(digest.Digest) (io.ReadCloser, error)
+
+	// CopyTo copies the content from the provided digest
+	// to the destination file path.
+	CopyTo(string, digest.Digest) error
```
stevvooe: This is an odd interface for providers to implement. Couldn't we hide this behind the `io.ReadCloser` by returning back an `os.File` directly?

Do you have an example implementation?

dmcgowan: Let me try keeping the interface the same and asserting for `*os.File` (or maybe has `Fd() uintptr`). I can probably just hide the use of `CopyFileRange` down in the apply rather than leaving it up the implementer to attempt. (It is a bit low level). The CopyFileRange interface does make it a bit trickier to use because of its requirement to have an offset, I can either assume an offset at the beginning in that case or read for the current offset.

dmcgowan:
Marked as work in progress, updated to not make interface changes

digests.go
```diff
@@ -10,26 +11,22 @@ import (
 
 // Digester produces a digest for a given read stream
 type Digester interface {
-	Digest(io.Reader) (digest.Digest, error)
+	Digest(ctxu.Context, io.Reader) (digest.Digest, error)
```
stevvooe: Why add a context here?

copy_unsupported.go
```diff
@@ -0,0 +1,12 @@
...
+
+package continuity
+
+import (
+	"io"
+	"os"
+)
+
+func fastCopy(dst *os.File, r io.Reader, size int64) (int64, error) {
+	return 0, nil
```
stevvooe: This should return an error declaring a lack of support.

PR #64, opened 2017-04-10
Separate windows and unix code

dmcgowan:
Move unix specific code to unix only build files. Add stub implementation for Windows where needed. Remove syscall importing from platform-agnostic files.

This isn't perfect but support for devices and symlinks on Windows is still unclear to me. Also does not appear that there is any equivalent of UID and GID to set.

### DIFF
--- a/resource_unix.go
+++ b/resource_unix.go
@@ -1,0 +1,37 @@
+// +build linux darwin
+
+package continuity
+
+import (
+	"fmt"
+	"os"
+	"syscall"
+)
+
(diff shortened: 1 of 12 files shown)


AkihiroSuda:
Can we add `GOOS=windows make` to `.travis.yml`?

@stevvooe please enable travis?

resource_unix.go
```diff
@@ -0,0 +1,37 @@
+// +build linux darwin
```
AkihiroSuda: `!windows`?

stevvooe:
LGTM

PR #66, opened 2017-04-30
Build on ppc64(le), s390x

sul3n3t:
These stubs will allow compilation on ppc64, ppc64le, and s390x architectures.

### DIFF
--- a/sysx/xattr.go
+++ b/sysx/xattr.go
@@ -2,10 +2,13 @@ package sysx
 
 import (
 	"bytes"
+	"fmt"
 	"syscall"
 )
 
 const defaultXattrBufferSize = 5
+
+var ErrNotSupported = fmt.Errorf("not supported")
 
 type listxattrFunc func(path string, dest []byte) (int, error)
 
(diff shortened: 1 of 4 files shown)


sul3n3t:
@stevvooe PTAL

stevvooe:
LGTM

These probably need support. cc @estesp

sul3n3t:
Thanks @stevvooe! Do you think this is mergeable or would you like to wait on @estesp feedback?

stevvooe:
@sul3n3t I'd like to get a little feedback to know if we need particular "road map" issues.

estesp:
A quick spot check seems to show these syscalls (the xattr family of calls) are available in current Golang implementation for ppc64le and s390x (for example: [https://github.com/golang/go/blob/master/src/syscall/zsyscall_linux_ppc64le.go](https://github.com/golang/go/blob/master/src/syscall/zsyscall_linux_ppc64le.go))

So; obviously this PR is one way to get a building project for those architectures, but seems the long term solution is an update to the generate scripts to add the pairs for `GOARCH`/`GOOS` to get actual proper generation for these syscalls.

PR #67, opened 2017-05-01
proto: improve comments

AkihiroSuda:

### DIFF
--- a/proto/manifest.proto
+++ b/proto/manifest.proto
@@ -24,8 +24,8 @@ message Resource {
 
     // user and group are not currently used but their field numbers have been
     // reserved for future use. As such, they are marked as deprecated.
-    string user = 4 [deprecated=true];
-    string group = 5 [deprecated=true];
+    string user = 4 [deprecated=true]; // "deprecated" stands for "reserved" here
+    string group = 5 [deprecated=true]; // "deprecated" stands for "reserved" here
 
     // Mode defines the file mode and permissions. We've used the same
     // bit-packing from Go's os package,
@@ -40,7 +40,9 @@ message Resource {
     uint64 size = 7;
 
     // Digest specifies the content digest of the target file. Only valid for
-    // regular files. The strings are formatted as <alg>:<digest hex bytes>.
+    // regular files. The strings are formatted in OCI style, i.e. <alg>:<encoded>.
+    // For detailed information about the format, please refer to OCI Image Spec:
+    // https://github.com/opencontainers/image-spec/blob/master/descriptor.md#digests-and-verification
     // The digests are sorted in lexical order and implementations may choose
     // which algorithms they prefer.
     repeated string digest = 8;
--- a/proto/manifest.pb.go
+++ b/proto/manifest.pb.go
@@ -71,7 +71,9 @@ type Resource struct {
 	// for regular files.
 	Size uint64 `protobuf:"varint,7,opt,name=size" json:"size,omitempty"`
 	// Digest specifies the content digest of the target file. Only valid for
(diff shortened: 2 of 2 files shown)


proto/manifest.proto
```diff
@@ -12,22 +12,26 @@ message Resource {
     // Path specifies the path from the bundle root. If more than one
     // path is present, the entry may represent a hardlink, rather than using
     // a link target. The path format is operating system specific.
+    // Path MAY be an absolute path, but it MUST be interpreted with the bundle root.
```
AkihiroSuda: current `manifest_test.go` uses relative path but `continuity build` uses absolute path.
Suggestion: can we prohibit absolute path for ease of robust implementation?

proto/manifest.proto
```diff
@@ -12,22 +12,26 @@ message Resource {
...
     // compatibility across different OS.
+    // On Unix, uid MUST be a positive decimal integer or an empty string.
     string uid = 2;
 
     // Gid specifies the group id for the resource. A string type is used for
     // compatibility across different OS.
+    // On Unix, gid MUST be a positive decimal integer or an empty string.
     string gid = 3;
 
+    // User and group specify the user name and the group name for the resource.
```
stevvooe: These fields aren't used, so this definition is meaningless.

proto/manifest.proto
```diff
@@ -12,22 +12,26 @@ message Resource {
...
     // a link target. The path format is operating system specific.
+    // Path MAY be an absolute path, but it MUST be interpreted with the bundle root.
     repeated string path = 1;
 
     // NOTE(stevvooe): Need to define clear precedence for user/group/uid/gid precedence.
 
     // Uid specifies the user id for the resource. A string type is used for
     // compatibility across different OS.
+    // On Unix, uid MUST be a positive decimal integer or an empty string.
     string uid = 2;
```
stevvooe: Let's just change these to integers. Windows aren't gonna use these, so they'll always be integers. Use `int64`.

AkihiroSuda: * not uint64?
* please elaborate your plan for windows?

stevvooe: > please elaborate your plan for windows?

You'll have to ask the microsoft folks, but they made the indication that we are going to use separate fields for that purpose. The only reason these are strings is due to the fact Windows has string user ids.

> not uint64?

Don't use integer types to enforce signedness.

AkihiroSuda: cc @jhowardmsft

jhowardmsft: @AkihiroSuda https://github.com/stevvooe/continuity/pull/65#discussion_r113073051

AkihiroSuda: thank you, will update soon

AkihiroSuda:
updated PR and opened https://github.com/containerd/continuity/pull/77 for unstringifying uid&gid

proto/manifest.proto
AkihiroSuda: @stevvooe WDYT?

stevvooe: Do we need to have a change to make this work, as well?

stevvooe: We should always be using relative paths for continuity.

AkihiroSuda: except symlink, right? (e.g. some executable can be a symlink to `/bin/true` or `/bin/false` on the absolute root)

stevvooe:
@AkihiroSuda Paths should be relative.

AkihiroSuda:
ok, I will open a PR in a few days

AkihiroSuda:
updated PR, and opened #82

stevvooe:
LGTM

PR #68, opened 2017-05-02
Require bash specifically; this does not work with a Posix shell

justincormack:


stevvooe:
LGTM

PR #73, opened 2017-05-04
Initial FreeBSD port

justincormack:
Currently stubs out extended attributes as these differ from the Linux
and Darwin ones and are not support in syscall package so need writing.
As extended attributes are rarely used on FreeBSD this is probably fine
for an initial port.

Fixes the other differences.


justincormack:
cc @dmcgowan

dmcgowan:
LGTM

Note that `SYS_FCHMODAT` is defined in `golang.org/x/sys/unix` but it is not generating `Fchmodat` yet. Will add this to the list of changes we need to push upstream.

PR #75, opened 2017-05-09
Provide a constant for ENODATA

justincormack:
There is no real standardisation between Unix versions as to whether
ENODATA or ENOATTR is used, so provide a standard constant in sysx that
can be used by consumers.

Linux returns ENODATA, while FreeBSD and Darwin return ENOATTR; to
confuse things ENODATA is defined on Darwin but is not returned for
extended attributes.

Applications should use `sysx.ENODATA` instead.

### DIFF
--- a/sysx/nodata_linux.go
+++ b/sysx/nodata_linux.go
@@ -1,0 +1,7 @@
+package sysx
+
+import (
+	"syscall"
+)
+
+const ENODATA = syscall.ENODATA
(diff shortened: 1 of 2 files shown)


justincormack:
May need to be rebased on #74 to pass CI

cc @dmcgowan

justincormack:
Oh ok, CI does not seem to check `go fmt`

stevvooe:
LGTM

Issue #63, opened 2017-04-10
Upstream `sysx` additions and use `golang.org/x/sys/unix`


dmcgowan:
We shouldn't be trying to maintain a system package and should only include changes which haven't been accepted upstream. The upstream package has much better support for a wider range of architectures which we could support.

PR #76, opened 2017-05-17
Add xattr definitions for power/s390x

tonistiigi:
@estesp 

Signed-off-by: Tonis Tiigi <tonistiigi@gmail.com>


stevvooe:
LGTM

@dmcgowan PTAL

estesp:
ping @tophj-ibm ; do you want do a quick look at this? Looks correct to me and is generated from the Go implementations, so I assume correct.

tophj-ibm:
Thanks @estesp and @tonistiigi !!

I just want to reiterate #63 in that the syscall in general isn't the greatest for a lot of reasons and should be switched to /x/sys ASAP. This is especially true for non-x86 architectures where a lot of calls haven't been well tested, and if an issue arrises it often takes an entire release cycle to get in. /rant 😄 

Having said that, I don't know of any issues with xattr, and I did check all of these calls against /x/sys just in case and I these look fine to me.

dmcgowan:
LGTM

justincormack:
@tophj-ibm the only constants here are the actual syscall numbers which are less likely to be wrong. But we could consider other ways of doing this.

PR #77, opened 2017-05-25
[breaking change] proto: uid, gid: string -> int64

AkihiroSuda:
For https://github.com/containerd/continuity/pull/67/files#r114387100


dmcgowan:
+1
We have discussed doing this for awhile as having usernames is never desired and we discovered having it be a string has no benefit for windows.

stevvooe:
LGTM

This is breaking.

PR #79, opened 2017-05-25
[WIP] Remote filesystem driver (Linux containers on Windows)

gupta-ak:
Added interface changes to the Driver interface to support manipulating Linux paths + files from Windows. The plan is that the Driver interface would be used to support operations like `docker cp`. Right now it's just interface changes; the actual implementation is stubbed out.

A branch with examples of how it would get used:
https://github.com/Microsoft/docker/tree/akagup/lcow-remotefs-impl


common/error.go
```diff
@@ -0,0 +1,8 @@
+package common
```
stevvooe: Please, no "common" packages. They become horrible.

stevvooe: :)

fsdriver/driver.go
```diff
@@ -35,19 +52,58 @@ type Driver interface {
...
-	// Abs(string) (string, error)
-	// Rel(base, target string) (string, error)
+	Join(pathName ...string) string
+	IsAbs(pathName string) bool
+	Rel(base, target string) (string, error)
+	Base(pathName string) string
+	Dir(pathName string) string
+	Clean(pathName string) string
+	Split(pathName string) (dir, file string)
+	Separator() byte
```
stevvooe: Does `path/filepath` export a type for this?

fsdriver/driver.go
```diff
@@ -35,19 +52,58 @@ type Driver interface {
 	// NOTE(stevvooe): We may want to actually include the path manipulation
 	// functions here, as well. They have been listed below to make the
 	// discovery process easier.
-
-	// Join(path ...string) string
-	// IsAbs(string) bool
-	// Abs(string) (string, error)
-	// Rel(base, target string) (string, error)
+	Join(pathName ...string) string
```
stevvooe: Please use `path` or `paths` for the argument names.

fsdriver/driver.go
```diff
@@ -35,19 +52,58 @@ type Driver interface {
...
+	// Abs(pathName string) (string, error)
 	// Walk(string, filepath.WalkFunc) error
 }
 
-func NewSystemDriver() (Driver, error) {
+// Unfortunately, os.File is a struct instead of an interface, an interface
+// has to be manually defined.
+var _ File = &os.File{}
+
+type File interface {
```
stevvooe: Does file actually need to be abstracted?

fsdriver/driver.go
```diff
@@ -1,9 +1,26 @@
...
+package fsdriver
 
 import (
 	"os"
+	"runtime"
+
+	"github.com/containerd/continuity/common"
+)
+
+type DriverType int
```
stevvooe: This type is unnecessary. Just instantiate them from different packages.

fsdriver/driver_basic.go
```diff
@@ -0,0 +1,94 @@
+package fsdriver
+
+import (
+	"os"
+	"path/filepath"
+)
+
+// basicDriver is a simple default implementation that sends calls out to the "os"
```
stevvooe: `localDriver`.

fsdriver/driver_basic_darwin.go
```diff
@@ -0,0 +1,11 @@
+package fsdriver
+
```
stevvooe: Need build tags here, but this also should apply for unices.

fsdriver/driver.go
```diff
@@ -1,9 +1,26 @@
-package continuity
+package fsdriver
```
stevvooe: `driver` is fine here.

fsdriver/driver.go
stevvooe: Also, are all these methods required?

fsdriver/driver_low.go
```diff
@@ -0,0 +1,100 @@
...
+
+	"github.com/containerd/continuity/common"
+	"github.com/containerd/continuity/devices"
+)
+
+// lowDriver is the Linux on Windows lowDriver that allows manipulating
+// files on a remote linux filesystem on Windows like a NFS client.
+// Right now, pretend it's a local file system. Later on, this can be
+// implemented through a NFS client, 9p client or another way.
+
```
stevvooe: Remove the space.

fsdriver/driver_low.go
```diff
@@ -0,0 +1,100 @@
...
+// replaced by the values from attr. If the operation fails to set any
+// attribute, those already applied will not be rolled back.
+func (d *lowDriver) Setxattr(path string, attrMap map[string][]byte) error {
+	return common.ErrNotSupported
+}
+
+// LGetxattr returns all of the extended attributes for the file at path p
+// not following symbolic links.
+func (d *lowDriver) LGetxattr(p string) (map[string][]byte, error) {
+	return nil, common.ErrNotSupported
```
stevvooe: Just don't implement these if they are not supported.

fsdriver/driver.go
stevvooe: Also, let's have a separate `PathDriver` that can be resolved.

stevvooe:
This is a great start! Comments in line.

fsdriver/driver.go
gupta-ak: There isn't an exported type. It's a constant. filepath.Separator == os.PathSeparator == '/' or '\\', so it's a byte currently. I am considering just to return a string for convenience since there is bunch of code that does string(filepath.Separator) in docker.

fsdriver/driver_basic_darwin.go
gupta-ak: hmm... I just moved the file from driver_darwin.go to fsdriver/driver_basic_darwin.go and changed the package name. Seems like driver_darwin.go can just be deleted from the master branch.

fsdriver/driver.go
gupta-ak: I personally don't need those methods implemented. I could just have driver.Open fail for the LOW driver. However, it seems weird to me that there is a dependency on *os.File. Since *os.File isn't an interface, every implementation would have to eventually call os.Open to compile.

stevvooe: What file operations do we actually need? `continuity` itself will not atomic writes. We may want to do more os-specific stuff, but it seems reasonable to get direct access to os-specific handle.

Maybe, we put this in another interface?

fsdriver/driver_basic_darwin.go
stevvooe: This should compile on darwin. Please don't remove support. This was probably already working, so adding build tags will break things. Should probably be called `_unix.go` or something like that.

gupta-ak:
After talking to Steve, we agreed that the implementation shouldn't be in continuity. So, I'm going to simply provide a PathDriver interface. I'm closing this pull request and then creating a new one with the changes.

stevvooe:
@gupta-ak Apart from the LCOW implementation here, the rest of this PR was a solid direction. In particular, the package organization was a fine direction.

Do you want to add that work to #80?

gupta-ak:
Yeah, I'll add the package organization back in.

PR #80, opened 2017-05-26
PathDriver interface to manipulate system paths

gupta-ak:
Removes implementation from #79 .  Adds a path driver interface and a basic implementation similar to driver.go.


gupta-ak:
@stevvooe I added the package changes back in.

conterrors/error.go
```diff
@@ -0,0 +1,10 @@
+package conterrors
```
stevvooe: Don't make a package for this: just define the errors where they are thrown and keep a helper function that handles the various types.

pathdriver/path_driver.go
```diff
@@ -0,0 +1,79 @@
...
+	IsAbs(path string) bool
+	Rel(base, target string) (string, error)
+	Base(path string) string
+	Dir(path string) string
+	Clean(path string) string
+	Split(path string) (dir, file string)
+	Separator() byte
+	Abs(path string) (string, error)
+	Walk(string, filepath.WalkFunc) error
+	NormalizePath(path string) string
```
stevvooe: Since this isn't a direct abstraction to filepath, make sure to document the expectations, or just call it `FromSlash`.

stevvooe:
LGTM

@dmcgowan PTAL

AkihiroSuda:
ping @dmcgowan

dmcgowan:
LGTM

PR #81, opened 2017-05-31
Remove duplicated function declared for darwin

dmcgowan:
Device info is already declared for unix and matches exactly the darwin definition. Continuity building for darwin is currently broken because of this.


crosbymichael:
LGTM

PR #82, opened 2017-06-07
[breaking change] proto: path: absolute -> relative

AkihiroSuda:
discussed in https://github.com/containerd/continuity/pull/67

### DIFF
--- a/context.go
+++ b/context.go
@@ -587,7 +587,7 @@ func (c *context) contain(p string) (string, error) {
 
 	// ZOMBIES(stevvooe): In certain cases, we may want to remap these to a
 	// "containment error", so the caller can decide what to do.
-	return filepath.Join("/", filepath.Clean(sanitized)), nil
+	return filepath.Clean(sanitized), nil
 }
 
 // digest returns the digest of the file at path p, relative to the root.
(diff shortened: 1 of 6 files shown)


dmcgowan:
I see the original statement about switching to relative paths, but not sure I agree. Unlike tars, continuity manifests are intended to describe a complete volume. When applied to a context directory, it is not intended that the manifest will merge with the existing system. On creation, it is not expected that only a subset of directories will be chosen or a random collection of files will be included, rather the context directory is intended to be preserved as is by the manifest. To this end, you can think of a manifest more as describing something that could be treated as a mount, that mount has its own root and that root will be treated relative to wherever it is mounted, but the mount has a root. By the same logic, the manifest has a root and it is appropriate to describe the files in that manifest as absolute paths with that root.

AkihiroSuda:
@stevvooe WDYT?

stevvooe:
@dmcgowan's argument is fairly convincing. It looks like the existing code treats everything as absolute paths, which seems to have been my intention at the time.

I'm sorry for sending you down this route.

What do we need to do to standardize on absolute paths?

AkihiroSuda:
sure, opened issue https://github.com/containerd/continuity/issues/84 toward standardizing absolute path spec

Issue #30, opened 2016-02-08
Support ctime and mtime


aaronlehmann:
Continuity should capture ctime (file creation timestamp) and mtime (file modification timestamp) in manifests, and by default, it should apply and verify these properties.

In case a particular application of Continuity needs to be insensitive to creation or modification times, it should be possible to use `--ignore-attr` at verify or apply time to exclude these attributes (see #27).

AkihiroSuda:
Seems dupe of #16

PR #88, opened 2017-07-19
proto: support timestamps

### LINKED ISSUE #16: Add mod time and creation time to manifest
In order to fully reproduce a source file system we should include the modification and creation times in the manifest.

AkihiroSuda:
Fix #16 #30 

Applier is not implemented yet

@stevvooe @dmcgowan @aaronlehmann @tonistiigi 


proto/manifest.proto
```diff
@@ -65,6 +67,13 @@ message Resource {
     // Ads stores one or more alternate data streams for the target resource.
     repeated ADSEntry ads = 13;
 
+    // Timestamps.
+    // btime (birthtime) is available on Darwin, FreeBSD, NetBSD, and Windows.
+    // When btime is unknown, it SHOULD be set to zero.
```
aaronlehmann: Why not omit the field instead of setting it to zero? I believe these fields are nullable.

resource.go
```diff
@@ -28,6 +31,11 @@ type Resource interface {
 
 	UID() int64
 	GID() int64
+
+	ATime() time.Time
+	MTime() time.Time
+	CTime() time.Time
+	BTime() time.Time
```
aaronlehmann: I think it's fine to use these abbreviated forms - they're pretty widely used. But it would be useful to have a comment explaining what each one means.

resource.go
```diff
@@ -108,6 +116,22 @@ func Merge(fs ...Resource) (Resource, error) {
 			return nil, fmt.Errorf("gid does not match: %v != %v", f.GID(), prototype.GID())
 		}
 
+		if f.ATime() != prototype.ATime() {
+			return nil, fmt.Errorf("atime does not match: %v != %v", f.ATime(), prototype.ATime())
```
aaronlehmann: I'm a little worried this will be racy. If a hardlinked file is accessed in between the `Stat`s, atime wouldn't match. The existing checks for things like mode and UID/GID are much less likely to change in the narrow window between two `Stat`s.

Should we avoid considering atime when merging hardlinks?

resource.go
```diff
@@ -503,13 +560,25 @@ func toProto(resource Resource) *pb.Resource {
 	return b
 }
 
+func timestampFromProto(ts *tspb.Timestamp) time.Time {
```
aaronlehmann: Let's return `time.Time{}` when `ts == nil`. This will clearly distinguish missing timestamps, for example from old manifests that don't implement timestamping, or ones generated on OSs that don't support every field. Otherwise we would return the unix epoch here.

context.go
```diff
@@ -190,6 +194,26 @@ func (c *context) Resource(p string, fi os.FileInfo) (Resource, error) {
 	return nil, ErrNotFound
 }
 
+func (c *context) verifyTimestamps(resource, target Resource) error {
+	unixZero := func(t time.Time) bool {
```
stevvooe: What's wrong with `t.IsZero()`?

resource.go
stevvooe: Well, it might be okay to follow the times library and just have `AccessTime`, `ModTime` and `ChangeTime`.

context.go
stevvooe: More interestingly, why do we care about unix time?

context.go
```diff
@@ -639,3 +669,18 @@ func (c *context) resolveXAttrs(fp string, fi os.FileInfo, base *resource) (map[
...
+func (c *context) resolveTimestamps(fi os.FileInfo, base *resource) {
+	ts := times.Get(fi)
+	base.atime = ts.AccessTime()
+	base.mtime = ts.ModTime()
+	base.ctime = ts.ChangeTime()
+	if ts.HasBirthTime() {
+		base.btime = ts.BirthTime()
+	} else {
+		// Note: time.Time{} stands for 0001-01-01 00:00:00 UTC
+		base.btime = time.Unix(0, 0)
```
stevvooe: I'm very worried about the redefinition of the zero time. This should really be `time.Time{}`.

stevvooe: Does that get serialized correctly?

resource.go
stevvooe: Just use the protobuf helper function here.

AkihiroSuda:
I'm not sure whether we should use 0001-epoch timestamp. i.e. `time.Time{}` rather than `time.Unix(0, 0)`

According to Wikipedia, most filesystems except FAT support Unix(1970)-epoch zero but not 0001-epoch zero: 

Filesystem | From | To
---| --- | ---
Ext4| December 14, 1901 |      May 10, 2446
Ext3| December 14, 1901 | January 18, 2038
Ext2| December 14, 1901 | January 18, 2038
XFS|  December 14, 1901 | January 18, 2038
HFS+| January 1, 1904     | February 6, 2040
NTFS| January 1, 1601     | May 28, 60056
FAT | January 1, 1980 | December 31, 2099 (December 31, 2107)

AkihiroSuda:
Any thought for timestamp epoch?

stevvooe:
> Any thought for timestamp epoch?

Use proper time types, ie `time.Time{}`.

proto/manifest.proto
```diff
@@ -65,6 +67,13 @@ message Resource {
...
+    google.protobuf.Timestamp atime = 14;
```
stevvooe: Make sure to document the fields.

resource.go
stevvooe: Document these. Please change to the expanded versions to match golang os package.

AkihiroSuda:
> Use proper time types, ie time.Time{}.

How to `apply` when it is before 19011214 (especially zero)

PR #90, opened 2017-07-20
Add solaris build support

epilatow:
since the last moby summit i've been working on porting containerd to
solaris.  to do this i need to make changes to this vendored component.

at the summit i was asked to try and err on the side of having multiple
smaller changesets rather than one large changeset.  to that end, this
pull request contains the smallest amount of changes necessary to get
this component to compile on solaris, to cross-compile for solaris on
linux, and to test solaris cross-compilation on linux via travis-ci.

it's worth mentioning that while these these changes allow solaris
cross-compilation on linux, the resultant binary isn't actually usable
on solaris.  this is because compilation of this component on solaris
requires the use of cgo, and we can't cross-compile solaris cgo code on
linux.  I chose to enable cross-compilation (even though the resultant
binary is broken) because that allows us to do testing via travis ci,
similar to what is currently being done for windows.

with these changes the component tests run successfully on solaris.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -12,6 +12,7 @@ env:
   - TRAVIS_GOOS=windows
   - TRAVIS_GOOS=linux
   - TRAVIS_GOOS=darwin
+  - TRAVIS_GOOS=solaris
 
 script:
 # TODO: vendor
@@ -19,3 +20,4 @@ script:
   - go get -t ./...
   - make build binaries
   - if [ "$GOOS" = "linux" ]; then make vet test; fi
+  - if [ "$GOOS" != "linux" ]; then make test-compile; fi
--- a/devices/devices_dummy.go
+++ b/devices/devices_dummy.go
@@ -1,0 +1,23 @@
+// +build solaris,!cgo
+
+//
+// Implementing the functions below requires cgo support.  Non-cgo stubs
+// versions are defined below to enable cross-compilation of source code
+// that depends on these functions, but the resultant cross-compiled
+// binaries cannot actually be used.  If the stub function(s) below are
+// actually invoked they will cause the calling process to exit.
+//
+
+package devices
+
+func getmajor(dev uint64) uint64 {
+	panic("getmajor() support requires cgo.")
+}
+
+func getminor(dev uint64) uint64 {
+	panic("getminor() support requires cgo.")
+}
+
+func makedev(major int, minor int) int {
+	panic("makedev() support requires cgo.")
+}
--- a/sysx/xattr_solaris.go
+++ b/sysx/xattr_solaris.go
@@ -4,9 +4,9 @@ import (
 	"errors"
 )
 
-// Initial stub version for FreeBSD. FreeBSD has a different
+// Initial stub version for Solaris. Solaris has a different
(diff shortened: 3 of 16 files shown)


.travis.yml
```diff
@@ -12,10 +12,12 @@ env:
...
   - TRAVIS_GOOS=darwin
+  - TRAVIS_GOOS=solaris
 
 script:
 # TODO: vendor
   - export GOOS=${TRAVIS_GOOS}
   - go get -t ./...
   - make build binaries
   - if [ "$GOOS" = "linux" ]; then make vet test; fi
+  - if [ "$GOOS" = "solaris" ]; then go test -c; fi
```
stevvooe: Create a target `make compile` test so the flags and env get setup consistently.

stevvooe: Also, this needs to be `go test -c ./...` to compile all packages.

devices/devices_dummy.go
```diff
@@ -0,0 +1,37 @@
...
+	"os"
+)
+
+func abort(name string) {
+	fmt.Fprintf(os.Stderr, "%s() support requires cgo.", name)
+	os.Exit(1)
+}
+
+func getmajor(dev uint64) uint64 {
+	abort("getmajor")
```
stevvooe: Just use a panic here.

sysx/xattr_solaris.go
```diff
@@ -0,0 +1,53 @@
...
+import (
+	"errors"
+)
+
+// Initial stub version for Solaris. Solaris has a different
+// syscall API from Darwin and Linux for extended attributes;
+// it is also not widely used. It is not exposed at all by the
+// Go syscall package, so we need to implement directly eventually.
+
+var unsupported error = errors.New("extended attributes unsupported on Solaris")
```
stevvooe: No need for type `error` here.

sysx/xattr_solaris.go
```diff
@@ -0,0 +1,53 @@
...
+}
+
+// Setxattr calls syscall setxattr
+func Setxattr(path string, attr string, data []byte, flags int) (err error) {
+	return unsupported
+}
+
+// Getxattr calls syscall getxattr
+func Getxattr(path, attr string) ([]byte, error) {
+	return []byte{}, nil
```
stevvooe: unsupported?

sysx/xattr_solaris.go
epilatow: this was copied from the freebsd version, so i'll fix it for both.

sysx/xattr_solaris.go
epilatow: this was copied from the freebsd version, so i'll commonize the functions and fix it for both.

.travis.yml
epilatow: turns out 'go test -c ...' doesn't allow you to specify multiple packages. so now i'm running 'go test -c <pkg>' for each package returned by 'go list ./...'

epilatow:
i've updated the PR with changes to address the previous comments.

stevvooe:
LGTM

@dmcgowan PTAL

dmcgowan:
LGTM

epilatow:
thanks

PR #91, opened 2017-07-25
[sysx] add ppc64le and s390x copyfilerange targets

tophj-ibm:
This adds ppc64le and s390x targets in the generate script in sysx, and generates them.

This should fix building / running tests for us on containerd.

The SYS_COPY_FILE_RANGE values were all taken from golang.org/x/sys.

Signed-off-by: Christopher Jones <tophj@linux.vnet.ibm.com>


dmcgowan:
Thanks, I forgot about this function. It has been upstreamed to `golang.org/x/sys/unix` and could be removed here, after containerd switches to no longer call it. Let's merge this now but I would prefer to see the update to containerd use the upstream package then delete it here.

dmcgowan:
LGTM

PR #92, opened 2017-08-09
sysx: remove custom implementation of CopyFileRange

tklauser:
golang.org/x/sys/unix provides the CopyFileRange syscall function on
Linux already, so there is no need for a custom wrapper. Now that
containerd switched to use the version from x/sys/unix (via
containerd/containerd#1312), there seem to be no remaining users of
sysx.CopyFileRange (according to [1]) and it can safely be removed.

  [1] https://godoc.org/github.com/containerd/continuity/sysx?importers

This should help #63 moving forward.


tklauser:
Pushed an update to mention #63 in the commit message, since it should also help with that.

AkihiroSuda:
SGTM

@stevvooe @dmcgowan PTAL?

dmcgowan:
LGTM

dmcgowan:
@AkihiroSuda I will take your SGTM as an LGTM 😄

PR #97, opened 2017-10-03
devices: use Mkdev, Major and Minor functions from golang.org/x/sys/unix

### LINKED ISSUE #96: devices: use helpers from x/sys/unix
See a similar change from containerd: https://github.com/containerd/containerd/pull/1580

tklauser:
Now that golang.org/x/sys/unix provides the Mkdev, Major and Minor
functions for every Unix-like OS, use them instead of the locally
defined versions.

Suggested by @stevvooe in containerd/containerd#1580.

Fixes #96


devices/devices_unix.go
```diff
@@ -14,42 +16,43 @@ func DeviceInfo(fi os.FileInfo) (uint64, uint64, error) {
 		return 0, 0, fmt.Errorf("cannot extract device from os.FileInfo")
 	}
 
-	return getmajor(sys.Rdev), getminor(sys.Rdev), nil
+	dev := uint64(sys.Rdev)
```
dmcgowan: `sys.Rdev` is already uint64, maybe just pass in directly to the functions

dmcgowan:
One nit, unless for some non-linux unix systems something different than `uint64` is used.

LGTM

stevvooe:
LGTM

@tklauser Thanks!

devices/devices_unix.go
tklauser: e.g. Darwin uses `int32`, FreeBSD uses `uint32`. Only Linux already uses `uint64` AFAICS. Thus, the type conversion is necessary.

tklauser:
> One nit, unless for some non-linux unix systems something different than uint64 is used.

Yes, most of the non-Linux ones use something different than `uint64` and the type conversion is needed there.

PR #98, opened 2017-11-14
vendor dependencies with vndr (https://github.com/LK4D4/vndr)

ijc:
Ran `vndr init`, sorted the resulting `vendor.conf` and committed the result.
    
Signed-off-by: Ian Campbell <ijc@docker.com>

In my first cut something regenerated `AUTHORS` for me, in my second go around that didn't happen so I forced it with `rm AUTHORS && make AUTHORS`.


estesp:
Looks like the travis config needs some love to not try and `vet` vendored deps

ijc:
Actually, I think all of `vet`, `build` and `test` in the `Makefile` could do with that. I'll look at it.

ijc:
Seems like the go version could be bumped too it's currently 1.7, 1.8 and tip, I proposed to add 1.9 and drop 1.7.

AkihiroSuda:
ping @stevvooe ptal

stevvooe:
LGTM

PR #103, opened 2018-02-09
Ensure rmdir gets set for every directory

dmcgowan:
Fixes a bug where rmdir was not getting set on a directory if the
previous change was also a directory.

This also needs to be cherry-picked into containerd 1.0.x


stevvooe:
LGTM

crosbymichael:
LGTM

PR #106, opened 2018-02-14
Export `copyFile`

cpuguy83:


AkihiroSuda:
cc @stevvooe @dmcgowan

dmcgowan:
Where is this planned on being used? I know we had some plans for this and possibly merging it with some functionality for buildkit's fsutil and Moby.

I am fine merging this and changing it later if we need to though.

LGTM

cpuguy83:
I was going to use this for something... what I don't remember... maybe it was in moby.

fs/copy.go
```diff
@@ -103,7 +103,9 @@ func copyDirectory(dst, src string, inodes map[uint64]string) error {
 	return nil
 }
 
-func copyFile(source, target string) error {
+// CopyFile copies the source file to the target.
+// The most efficient means of copying is used for the platform.
+func CopyFile(source, target string) error {
```
stevvooe: We usually use LHS notation for the args in this package.

Should be `CopyFile(target, source string)`

cpuguy83: Fixed

PR #108, opened 2018-02-23
sysx: use xattr functions from x/sys/unix

tklauser:
Use the xattr functions from package golang.org/x/sys/unix instead of
syscall. Also use its L*xattr functions instead of locally duplicating
them.


dmcgowan:
Which functions are still missing from `golang.org/x/sys/unix`? The goal of `sysx` is to eventually go away and be replaced by `golang.org/x/sys/unix`.

tklauser:
AFAICS, the `*xattr` functions for `darwin` are missing from `x/sys/unix`. I'll send a CL for those and update this PR afterwards. Plus, there are a few wrappers in `sysx`, e.g. `getxattrAll` or `listxattrAll`. I'm not sure whether we should add these to `x/sys/unix` though.

dmcgowan:
LGTM

PR #109, opened 2018-03-04
Add OpenBSD support for fs subpackage

nogoegst:


fs/stat_openbsd.go
```diff
@@ -0,0 +1,28 @@
+// +build openbsd
```
dnephin: Isn't this identical to `stat_linux.go` ? could we use the same file with build tags?

nogoegst: oh wow, nice catch!
`stat_linux.go` has to be renamed though. what name should it have?

nogoegst:
hey there again!
I made corresponding changes and squashed all the things :)

nogoegst:
gentle ping :)

nogoegst:
@dnephin ping :)

nogoegst:
I have no idea why CI failed here in one case.

nogoegst:
It seems it was a CI issue last time.

PR #113, opened 2018-05-09
Resolve context root on context.Walk when root is symlink

darstahl:
When the root of a context is a symlink, Walk does not walk the contents of the directory as expected due to the implementation in Golang. This expands the root of a context to the target of the link prior to the walk in order to correctly walk the contents of the context.

Readlink was forked on Windows as there is a bug in Golang that prevents Volume GUID paths of the form `\??\Volume` from being a valid target of a link. This type of path is necessary for mounted VHDs of Windows container filesystems.

Signed-off-by: Darren Stahl <darst@microsoft.com>


context.go
```diff
@@ -572,8 +572,16 @@ func (c *context) Apply(resource Resource) error {
...
+	root := c.root
+	fi, err := c.driver.Lstat(c.root)
+	if err == nil && fi.Mode()&os.ModeSymlink != 0 {
+		root, err = c.driver.Readlink(c.root)
+		if err != nil {
+			return err
+		}
+	}
+	return c.pathDriver.Walk(root, func(p string, fi os.FileInfo, err error) error {
+		contained, err := c.containWithRoot(p, root)
```
dmcgowan: Is the above logic duplicated in container with root?

dmcgowan:
Are you opening up an upstream change to fix the `ReadLink` issue? `sysx` is _intended_ to only have temporary changes before they are upstreamed to https://github.com/golang/sys.

context.go
darrenstahlmsft: It was, but shouldn't have been. Removed the duplicated logic. `containWithRoot` now uses the root as given with no modifications.

darrenstahlmsft:
Yes, the changes will be submitted to Go's Syscall package. The current state of https://github.com/golang/sys syscall package is not up to date with Go master (which contains the patch that broke this), so I'm not sure if this change will land in sys or syscall or both. It is intended to be upstreamed though, and this is just a temporary fork.

dmcgowan:
LGTM

kolyshkin:
> Yes, the changes will be submitted to Go's Syscall package.

@darstahl can you please point out to those? I did a quick look and can't seem to find any.

kolyshkin:
Found myself re-reading this today :) If anyone knows if the Windows' readlink bug is fixed in golang, please share the news

kolyshkin:
Guess what, I found the issue (https://github.com/golang/go/issues/30463) and the fix (https://go-review.googlesource.com/c/go/+/164201, very similar to what we have in here).

Will follow up with the clean patch

PR #114, opened 2018-05-16
Fix copy_file_range usage for files > 2GB

cpuguy83:
copy_file_range (or any Linux syscall) will only ever copy 2GB at a
time. In this case there is no error returned from the system call.

This fix uses copy_file_range in a loop until either 0 bytes are copied,
or the desired amount has been copied (st.Size()).

### DIFF
--- a/fs/copy_linux.go
+++ b/fs/copy_linux.go
@@ -49,20 +49,26 @@ func copyFileContent(dst, src *os.File) error {
 		return errors.Wrap(err, "unable to stat source")
 	}
 
-	n, err := unix.CopyFileRange(int(src.Fd()), nil, int(dst.Fd()), nil, int(st.Size()), 0)
-	if err != nil {
-		if err != unix.ENOSYS && err != unix.EXDEV {
-			return errors.Wrap(err, "copy file range failed")
+	size := st.Size()
+	first := true
+	srcFd := int(src.Fd())
+	dstFd := int(dst.Fd())
+
+	for size > 0 {
+		n, err := unix.CopyFileRange(srcFd, nil, dstFd, nil, int(size), 0)
+		if err != nil {
+			if (err != unix.ENOSYS && err != unix.EXDEV) || !first {
+				return errors.Wrap(err, "copy file range failed")
+			}
+
+			buf := bufferPool.Get().(*[]byte)
+			_, err = io.CopyBuffer(dst, src, *buf)
+			bufferPool.Put(buf)
+			return errors.Wrap(err, "userspace copy failed")
 		}
 
-		buf := bufferPool.Get().(*[]byte)
-		_, err = io.CopyBuffer(dst, src, *buf)
-		bufferPool.Put(buf)
-		return err
-	}
-
-	if int64(n) != st.Size() {
-		return errors.Wrapf(err, "short copy: %d of %d", int64(n), st.Size())
+		first = false
+		size -= int64(n)
 	}
 
 	return nil
--- a/fs/copy_test.go
+++ b/fs/copy_test.go
@@ -1,11 +1,11 @@
 package fs
 
 import (
+	_ "crypto/sha256"
+	"io"
 	"io/ioutil"
 	"os"
 	"testing"
-
-	_ "crypto/sha256"
 
 	"github.com/containerd/continuity/fs/fstest"
 	"github.com/pkg/errors"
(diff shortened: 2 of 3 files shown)


cpuguy83:
ping @stevvooe @dmcgowan @AkihiroSuda @Random-Liu

This is a nasty gotcha. Considering this a P0 for moby/moby.
I suspect cri-containerd volume copying and containerd's image unpack is broken for any directories that contain files > 2GB.

fs/copy_linux.go
```diff
@@ -49,22 +49,30 @@ func copyFileContent(dst, src *os.File) error {
...
+	first := true
+	for written < size {
+		n, err := unix.CopyFileRange(int(src.Fd()), nil, int(dst.Fd()), nil, int(size-written), 0)
+		if err != nil {
+			if (err != unix.ENOSYS && err != unix.EXDEV) || !first {
+				return errors.Wrap(err, "copy file range failed")
+			}
 
-	if int64(n) != st.Size() {
-		return errors.Wrapf(err, "short copy: %d of %d", int64(n), st.Size())
```
cpuguy83: Note that this `errors.Wrap` is incorrect and always returns nil.

fs/copy_linux.go
```diff
@@ -49,22 +49,30 @@ func copyFileContent(dst, src *os.File) error {
...
-		}
+	var written int64
+	size := st.Size()
 
-		buf := bufferPool.Get().(*[]byte)
-		_, err = io.CopyBuffer(dst, src, *buf)
-		bufferPool.Put(buf)
-		return err
-	}
+	first := true
```
justincormack: you arent setting this to `false` later

cpuguy83: Fixed. It's probably useless, but figured why not have the extra protection.

fs/copy_linux.go
```diff
@@ -49,22 +49,31 @@ func copyFileContent(dst, src *os.File) error {
...
+			bufferPool.Put(buf)
+			return errors.Wrap(err, "userspace copy failed")
+		}
+		if n == 0 {
+			break
+		}
+		first = false
+		written += int64(n)
+	}
+	if written != size {
```
justincormack: this surely cant happen?

fs/copy_linux.go
```diff
@@ -49,22 +49,31 @@ func copyFileContent(dst, src *os.File) error {
...
+	size := st.Size()
 
-		buf := bufferPool.Get().(*[]byte)
-		_, err = io.CopyBuffer(dst, src, *buf)
-		bufferPool.Put(buf)
-		return err
-	}
+	first := true
+	for written < size {
+		n, err := unix.CopyFileRange(int(src.Fd()), nil, int(dst.Fd()), nil, int(size-written), 0)
```
justincormack: you shouldnt cast `size-written` to int, it could be larger. you will need a check for is it >2G and then write a smaller amount given the length arg is an int.

fs/copy_linux.go
cpuguy83: i.e., I guess only if there was a bug in the kernel in the go syscall implementation.

justincormack: its not really worth checking for random kernel bugs

fs/copy_linux.go
justincormack: (the syscall doesnt take an int here so in C you would be able to pass a 64 bit value)

cpuguy83: Fixed.

fs/copy_linux.go
```diff
@@ -43,28 +43,41 @@ func copyFileInfo(fi os.FileInfo, name string) error {
 	return nil
 }
 
+const maxCopySize = 2147483647 // max 32bit integer
```
crosbymichael: Could we just use `math.MaxInt32` here?

cpuguy83: Certainly. Didn't want to add the import just to get one number.

cpuguy83: Updated

fs/copy_linux.go
```diff
@@ -49,22 +50,33 @@ func copyFileContent(dst, src *os.File) error {
...
-	if err != nil {
-		if err != unix.ENOSYS && err != unix.EXDEV {
-			return errors.Wrap(err, "copy file range failed")
+	var written int64
+	size := st.Size()
+	first := true
+
+	for written < size {
+		var desired int
+		if size-written > math.MaxInt32 {
```
kolyshkin: this `size-written` looks ugly without spaces, but apparently it's gofmt way. So I'd suggest to rewrite it as
```go
desired := size - written
if desired > math.MaxInt32 {
	desired=int(math.MaxInt32)
}
```

kolyshkin: OK in fact I think we should not do something like this instead:

```golang
sfd := int(src.Fd())
dfd := int(dst.Fd())
size := st.Size()
for size > 0 {
		n, err := unix.CopyFileRange(sfd, nil, dfd, nil, int(size), 0)
		if err != nil {
				// ...
		}
		size -= n
}
```

This way we don't have any knowledge of how big the size can be, but just relying on the fact that the syscall returns the number of bytes actually copied.

Note that in here `int(size)` can indeed be negative, but it is internally used as `size_t`, ie an unsigned type.

cpuguy83: Yeah, I was wondering why it didn't fail before when I did this. This makes sense.

Random-Liu:
LGTM. Will update the vendor once this is merged.

fs/copy_linux.go
cpuguy83: Done.

fs/copy_test.go
```diff
@@ -46,6 +46,45 @@ func TestCopyDirectoryWithLocalSymlink(t *testing.T) {
 	}
 }
 
+// TestCopyWithLargeFile tests copying a file whose size is > 32bit.
```
kolyshkin: Maybe `size > 2^32 bytes`

cpuguy83: Done.

fs/copy_test.go
```diff
@@ -46,6 +46,45 @@ func TestCopyDirectoryWithLocalSymlink(t *testing.T) {
 	}
 }
 
+// TestCopyWithLargeFile tests copying a file whose size > 2^32 bytes.
+func TestCopyWithLargeFile(t *testing.T) {
```
dmcgowan: I'll consider this for a follow up, but skipping on `testing.Short()` may be a good idea

fs/copy_test.go
```diff
@@ -46,6 +46,45 @@ func TestCopyDirectoryWithLocalSymlink(t *testing.T) {
...
+	dataR, dataW := io.Pipe()
```
dmcgowan: I think a pseudorandom limit reader would make it cleaner

dmcgowan:
Some comments on test, I will address those in a follow up. Thanks for getting this in.

LGTM

PR #115, opened 2018-05-16
Make sure travis tests on latest go version.

cpuguy83:


.travis.yml
```diff
@@ -4,6 +4,8 @@ sudo: required
 go:
   - 1.8.x
   - 1.9.x
+  - 1.10.x
+  - 1.x
```
estesp: what does `1.x` do? And I would think we can drop `1.8.x`

cpuguy83: 1.x runs on the latest `1.x` line. Really it is currently redundant with 1.10.x, but will work when 1.11 is released.

dmcgowan:
I am ok keeping `1.8.x` for now, the tests are currently pretty fast.

LGTM

PR #116, opened 2018-05-21
Update TestCopyWithLargeFile

dmcgowan:
Add helper to create random files of any size and skip large file copy when short flag is given.


cpuguy83:
LGTM

PR #117, opened 2018-05-23
Fix bug in multiple calls to file applier

dmcgowan:
Ensure that file appliers create the same content every time they are called.


dmcgowan:
This was manifesting in the containerd tests and not here because container uses a base set of applier which are used in multiple tests. We could probably do something similar here, but not sure where that would fit since it would be testing the appliers.

crosbymichael:
LGTM

PR #118, opened 2018-06-22
 Add mechanisms to be compatible with Moby's copy library

sargun:
Make it so we can be used by Moby (easily)

### DIFF
--- a/fs/copy.go
+++ b/fs/copy.go
@@ -9,6 +9,16 @@ import (
 	"github.com/pkg/errors"
 )
 
+// Mode indicates whether to use hardlink or copy content
+type Mode int
+
+const (
+	// Content creates a new file, and copies the content of the file
+	Content Mode = iota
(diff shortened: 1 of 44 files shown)


fs/copy.go
```diff
@@ -9,6 +9,16 @@ import (
...
 )
 
+// Mode indicates whether to use hardlink or copy content
+type Mode int
+
+const (
+	// Content creates a new file, and copies the content of the file
+	Content Mode = iota
+	// Hardlink creates a new hardlink to the existing file
+	Hardlink
```
dmcgowan: Where is this used other than in one of moby's graph drivers? Moby specific changes should be kept in that repo, this project is focused on cleaner interfaces to support containerd and future fs metadata formats.

cpuguy83:
Closing since it seems like this change is not desirable here.

PR #120, opened 2018-07-11
Fix Lchmod() for Linux+Go1.11

kolyshkin:
This fixes the bug that Go 1.11 uncovered (rendering driver.Lchmod() useless as it always returns an error now. Found during moby rebase to go-1.11beta1 ( https://github.com/moby/moby/pull/37358#issuecomment-404032896).

This set also optimizes and cleans up some things here and there. Please see individual commit messages for all the details.

The initial 3 commits are preparations:
* [PATCH 1/5] sysx/xattr_darwin.go: rm duplicate Fchmodat def
* [PATCH 2/5] sysx.Fchmodat(): ~~unify~~ remove implementations
* [PATCH 3/5] Lchmod(): simplify and optimize

The 4th commit is the fix itself:
* [PATCH 4/5] Lchmod(): fix for Linux/Go 1.11

The last one is a small post-fix followup:
* [PATCH 5/5] context.Apply: no need to skip chmod on symlinks


driver/lchmod_linux.go
```diff
@@ -0,0 +1,19 @@
...
+
+	"github.com/containerd/continuity/sysx"
+)
+
+// Lchmod changes the mode of a file not following symlinks.
+func (d *driver) Lchmod(path string, mode os.FileMode) error {
+	// On Linux, file mode is not supported for symlinks,
+	// and fchmodat() does not support AT_SYMLINK_NOFOLLOW,
+	// so symlinks need to be skipped entirely.
+	if st, err := os.Stat(path); err == nil && st.Mode()&os.ModeSymlink != 0 {
```
thaJeztah: (Perhaps overthinking it, but leaving this here because I learned to never "assume" things 😅)

From https://go-review.googlesource.com/c/go/+/118658

```
Currently Linux' fchmodat(2) syscall implementation doesn't support the
flags parameter (though it might in future versions [1]). Fchmodat in
the syscall package takes the parameter and (wrongly) passes it on to the
syscall which will ignore it.

According to the POSIX.1-2008 manual page [2], AT_SYMLINK_NOFOLLOW is
the only valid value for the flags parameter and EOPNOTSUPP should be
returned in case changing the mode of a symbolic link is not supported
by the underlying system. EINVAL should be returned for any other value
of the flags parameter.

  [1] https://patchwork.kernel.org/patch/9596301/
  [2] http://pubs.opengroup.org/onlinepubs/9699919799/functions/chmod.html
```

That patch mentions the possibility that in future, other options may actually
become active.

Instead of _skipping_ `sysx.Fchmodat()` for symlinks, should we always run it, but check if the returned error was an `EOPNOTSUPP`, _then_ return `nil` if the path was a symlink?

kolyshkin: Well, one problem is `syscall.Fchmodat()` from golang < 1.11 does not return the error (same for unix.Fchmodat() from golang.org/x/sys/unix before https://github.com/golang/sys/commit/c23410a886927bab8ca5e80b08af6a56faeb330d which only landed about a year ago), instead changing the mode for the file that a symlink points to (bad bad bad!). So we can't rely on it returning the proper error, and from the correctness/security standpoint it makes sense to avoid the call.

Other things is, Linux won't support mode for symlinks, ever (I mean, I can't speak for Linus, but it looks like it). The patch being referred to above (https://patchwork.kernel.org/patch/9596301/) only adds flags argument to fchmodat(), for the sake of returning an error in case flags are specified. Even that one didn't made it to the kernel, and it seems to be the 3rd attempt, previous two happened in 2015 (https://groups.google.com/d/msg/linux.kernel/Z2z3_MR2aM8/mFz04xQhZEIJ) and 2014 (https://groups.google.com/d/msg/linux.kernel/8lnkJ-naNRI/QvkfwNih6rcJ).

kolyshkin: Please ignore the previous comment; while correct it misses the biggest issue, that is, EOPNOTSUPP is returned _always_, not just for the symlinks. In other words, calling `Fchmodat()` with flags != 0 is useless, it does nothing but returns an error, and it happens for all files, not just symlinks.

thaJeztah: Thanks for doing the digging; looks like the current approach is best then!

driver/lchmod_unix.go
```diff
@@ -0,0 +1,14 @@
...
+
+import (
+	"os"
+
+	"github.com/containerd/continuity/sysx"
+)
+
+// Lchmod changes the mode of a file not following symlinks.
+func (d *driver) Lchmod(path string, mode os.FileMode) error {
+	return sysx.Fchmodat(sysx.AT_FDCWD, path, uint32(mode), sysx.AT_SYMLINK_NOFOLLOW)
```
dmcgowan: Can we use `golang.org/x/sys/unix` directly here? The `sysx` package is only intended to supplement which doesn't exist in that package, not wrap it

driver/lchmod_linux.go
```diff
@@ -0,0 +1,19 @@
...
+		return nil
+	}
```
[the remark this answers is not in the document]
kolyshkin: This was my original approach, but then I realized I am changing the behavior that way (as previously the error (like ENOENT or EACCESS) was returned from `Fchmodat()`).

This code assumes that if `path` can't be `stat()` it is definitely not a symlink.

driver/lchmod_unix.go
kolyshkin: Sure; I can remove the `Fchmodat()` from `sysx` entirely (hoping it does not have any external users).

driver/lchmod_linux.go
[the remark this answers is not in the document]
dmcgowan: ahh, because the `os` package will wrap the error?

driver/lchmod_unix.go
kolyshkin: Updated, PTAL

kolyshkin:
Patchset updated to remove sysx.Fchmodat() entirely and use x/sys/unix directly as suggested by @dmcgowan. This looks way better.

driver/lchmod_linux.go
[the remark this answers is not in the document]
kolyshkin: For one thing, yes. I didn't want to inadvertently open a can of worms.

In general it's a problem that is out of scope for this PR. The problem is, some errors returned by continuity come from high-level Go functions (like the ones from `os`) and are of `PathError` type, some come from low-level (like `x/sys/unix` or `syscall`) and are not wrapped. The most user-visible difference is high-level ones contain file names and in general provide more clue about what is going on.

Based on the above, it would be good to review and unify errors returned (I am in favor of `os.PathError{}`). But again it's out of scope for this PR.

dmcgowan:
LGTM

crosbymichael:
LGTM

PR #122, opened 2018-07-15
Fix copy_file_range usage for files > 2GB on 32-bit archs

poizan42:
So this turned out to actually have nothing to do with MIPS, but just 32-bit architectures in general.

---
On mips copy_file_range actually fails with EINVAL for sizes over 2GB.

This fix ensures that we are never trying to copy more than 2GB at a time and at the same time avoids overflows when the file is larger than 4GB.

Note that when I tried building in a qemu vm I had to increase the test timeout because it took more than 10 minutes to run TestCopyWithLargeFile (in fs).

### DIFF
--- a/fs/copy_linux.go
+++ b/fs/copy_linux.go
@@ -59,6 +59,8 @@ func copyFileInfo(fi os.FileInfo, name string) error {
 	return nil
 }
 
+const maxSSizeT = int64(^uint(0) >> 1)
+
 func copyFileContent(dst, src *os.File) error {
 	st, err := src.Stat()
 	if err != nil {
@@ -71,7 +73,16 @@ func copyFileContent(dst, src *os.File) error {
 	dstFd := int(dst.Fd())
 
 	for size > 0 {
-		n, err := unix.CopyFileRange(srcFd, nil, dstFd, nil, int(size), 0)
+        // Ensure that we are never trying to copy more than SSIZE_MAX at a
+        // time and at the same time avoids overflows when the file is larger
+        // than 4GB on 32-bit systems.
+		var copySize int
+		if size > maxSSizeT {
+			copySize = int(maxSSizeT)
+		} else {
+			copySize = int(size)
+		}
+		n, err := unix.CopyFileRange(srcFd, nil, dstFd, nil, copySize, 0)
 		if err != nil {
 			if (err != unix.ENOSYS && err != unix.EXDEV) || !first {
 				return errors.Wrap(err, "copy file range failed")


fs/copy_linux.go
```diff
@@ -55,7 +56,13 @@ func copyFileContent(dst, src *os.File) error {
 	dstFd := int(dst.Fd())
 
 	for size > 0 {
-		n, err := unix.CopyFileRange(srcFd, nil, dstFd, nil, int(size), 0)
+		var copySize int
+		if size > int64(math.MaxInt32) {
```
dmcgowan: Is this a limitation on all platforms?

poizan42: According to @cpuguy83's comment in https://github.com/containerd/continuity/commit/afba265aa60a6dab3bf3a7809b2e6e9a95704b04 it never copies more than 2GB even when it doesn't fail. Actually looking at the kernel source it looks like it is actually up to the fs implementation. That said I don't think there's going to exist any storage in the near future where one syscall per 2GB data is going to be a non-negligible overhead.

Now that I was looking at the kernel it actually seems like sizes larger than 2GB won't work on any 32-bit system, see https://github.com/torvalds/linux/blob/30b06abfb92bfd5f9b63ea6a2ffb0bd905d1a6da/fs/read_write.c#L372, in which case the commit message here is misleading (is the travis build only testing on amd64? That might be a problem).

dmcgowan: > That said I don't think there's going to exist any storage in the near future where one syscall per 2GB data is going to be a non-negligible overhead.

We added support for this call to make use of shared extents on BTRFS and XFS which can perform these copies very fast since there is no byte copying, hence my wondering of why this limit would have any relevance on such a system. I guess what I want to understand is what is `int(size)` doing on these 32 bit systems when a large value is given? Perhaps we need a smarter truncation function than just casting and if MIPS has specific logic (such as 64 bit but limits 32 bits with this syscall) we can have a MIPS specific implementation which does a 64->32 bit truncation.

poizan42: Well int(size) is truncating when int is 32-bit which causes it to just wraparound. But the documentation for copy_file_range claims that it takes a size_t which should be uint in go, but it would work anyways because it just gets reinterpreted as unsigned in the kernel. But then again in the kernel it is actually checked that it doesn't overflow an ssize_t (which doesn't seems to be explicitly documented, but the return value is ssize_t), so there is no way to exceed 2GB at a time on 32-bit systems.

What happens with respect to shared extents larger than 2GB I do not know, but hopefully the filesystems are smart enough to detect this. Or it might not even be an issue because it might actually be implemented as a number of smaller extents right after each other. Anyways the limitation on the syscall is what it is.

Just to verify I tried building on x86 (i.e. 32-bit) and the test also fails with EINVAL. So at least we need to ensure that the size given never exceeds SSIZE_MAX.

dmcgowan: Could we change the logic then to detect the truncation then

```
copySize := int(size)
if size != int64(copySize) && size > int64(math.MaxInt32) {
	copySize = int(math.MaxInt32)
}
```

What I don't understand is how can `int(size)` be truncating yet `int(size) > int(math.MaxInt32)`. I assume `int(size) < 0` cannot be true even with wraparound.

poizan42: int(size) > int(math.MaxInt32) can only be true on 64-bit archs where int is 64-bits, but anyways I changed it to explicitly check against the max value of int which corresponds to SSIZE_MAX.

poizan42:
I should remember to actually include the updated file...

dmcgowan:
LGTM

cpuguy83:
Are we sure this fixes it? IIRC the way the sys call gets made this ends up converting to 2GB anyway.

poizan42:
@cpuguy83 As far as I can tell it is up to the fs implementation. Watch for yourself, the syscall entrypoint is [here]( https://github.com/torvalds/linux/blob/30b06abfb92bfd5f9b63ea6a2ffb0bd905d1a6da/fs/read_write.c#L1627) which calls [vfs_copy_file_range](https://github.com/torvalds/linux/blob/30b06abfb92bfd5f9b63ea6a2ffb0bd905d1a6da/fs/read_write.c#L1549) with copying in/out of the offsets. It then verifies the buffer/length with [rw_verify_area](https://github.com/torvalds/linux/blob/30b06abfb92bfd5f9b63ea6a2ffb0bd905d1a6da/fs/read_write.c#L365) which checks that it does not exceed max ssize_t (which is what caused this to fail on 32-bit systems).

So on 64-bit systems where ssize_t is int64 this succeeds. Then it uses clone_file_range if the fs supports that and if not it tries copy_file_range. It is only if the fs supports neither that we hit the [do_splice_direct call](https://github.com/torvalds/linux/blob/30b06abfb92bfd5f9b63ea6a2ffb0bd905d1a6da/fs/read_write.c#L1607) with a max size of [MAX_RW_COUNT](https://github.com/torvalds/linux/blob/ea5f39f2f994e6fb8cb8d0304aa5f422ae3bbf83/include/linux/fs.h#L2225) which limits the size to a bit below 2GB (but doesn't fail in that case if the size is larger).

AkihiroSuda:
What's current status? Is this mergable?
@cpuguy83 @dmcgowan @stevvooe

cpuguy83:
Should be. I don't have write access here.

estesp:
@poizan42 can you add a comment per @AkihiroSuda's request? Maybe something straight from your commit at the top of the `if size > 0` block?

Something like: *Ensure that we are never trying to copy more than SSIZE_MAX at a time and at the same time avoids overflows when the file is larger than 4GB on 32-bit systems.*

This looks worth getting in; happy to merge after getting a comment added.

poizan42:
CI fails with

     1f02a3c - FAIL - commit subject exceeds 90 characters

Isn't the "commit subject" the first line of the description? It is only 59 characters long.

estesp:
@poizan42 in `git`, until it finds a blank line, everything up to that point is the "subject"/title:
*The text up to the first blank line in a commit message is treated as the commit title, and that title is used throughout Git...*

You need a blank line after your 59 character title to fix the CI check.

poizan42:
This should be ready to be merged now

PR #123, opened 2018-08-27
driver/{Mknod,Mkfifo,Lchmod}: return PathError

kolyshkin:
In case any of these methods fail, make sure to return an error wrapped
in `os.PathError`, to be in line with the other methods (most of which are
wrappers for `os.Foo()` calls, which return `os.PathError` or `os.LinkError`).


dmcgowan:
LGTM

PR #127, opened 2018-09-11
testutil: sync with containerd

AkihiroSuda:
Signed-off-by: Akihiro Suda <suda.akihiro@lab.ntt.co.jp>

Port over https://github.com/containerd/containerd/blob/master/pkg/testutil/loopback_linux.go

After this PR gets merged we can remove the corresponding code from containerd and vendor continuity

@stevvooe @dmcgowan

PR #128, opened 2018-09-11
 fs: add TestCopyReflinkWithXFS

AkihiroSuda:
This PR makes sure `CopyFile()` supports reflink on XFS.

This PR also covers https://github.com/containerd/continuity/pull/127

### DIFF
--- a/fs/copy_linux_test.go
+++ b/fs/copy_linux_test.go
@@ -19,6 +19,9 @@
 package fs
 
 import (
+	"io"
+	"path/filepath"
+	"math/rand"
 	"io/ioutil"
 	"os"
 	"os/exec"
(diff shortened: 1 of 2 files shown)


fs/copy_linux_test.go
```diff
@@ -0,0 +1,79 @@
...
+		loopback.Close()
+	}()
+
+	aPath := filepath.Join(mnt, "a")
+	aSize := int64(100 << 20) // 100MB
+	a, err := os.Create(aPath)
+	if err != nil {
+		t.Fatal(err)
+	}
+	if _, err := io.CopyN(a, rand.Reader, aSize); err != nil {
```
dmcgowan: Don't use crypto rand for this

testutil/loopback_linux.go
```diff
@@ -49,5 +69,47 @@ func NewLoopback(size int64) (string, func() error, error) {
...
+// SoftSize returns st_size
+func (l *Loopback) SoftSize() (int64, error) {
+	st, err := os.Stat(l.File)
+	if err != nil {
+		return 0, err
+	}
+	return st.Size(), nil
+}
+
+// HardSize returns st_blocks * 512
```
dmcgowan: Can you link to somewhere that justifies 512 as a hardcoded value

fs/copy_linux_test.go
AkihiroSuda: What should I do instead?

testutil/loopback_linux.go
AkihiroSuda: http://man7.org/linux/man-pages/man2/fstat.2.html

> st_blocks
>  This field indicates the number of blocks allocated to the file, in 512-byte units.

fs/copy_linux_test.go
dmcgowan: `math/rand` should work just fine here and be much faster

fs/copy_linux_test.go
```diff
@@ -0,0 +1,81 @@
+// +build linux
```
dmcgowan: The file header checking was added, that needs to be added here to pass CI

dmcgowan:
LGTM after header added to make CI hapy

AkihiroSuda:
updated

PR #134, opened 2018-09-20
Remove unreachable block in fs path cleanup

dmcgowan:
The block checking for the root as a prefix of an absolute link is unnecessary since the end path always gets joined with the root. In the case where the symlink path is prefixed with the root, this should be considered coincidental and the unit tests already show that the root gets correctly applied to absolute links. 

Additionally inside this block there is a bug causing in incorrect truncation of the path which would
only manifest itself in this rare coincidental case.

ping @tonistiigi


tonistiigi:
LGTM

PR #136, opened 2018-10-01
split testutil/loopback_linux.go to another pkg

AkihiroSuda:
containerd/containerd#2694 is going to import both `containerd/pkg/testutil` and
`continuity/testutil` for deduplicating `NewLoopback()`.

However, as the both pkg defines `test.root` BoolVar in `init()`, the test panics.

This commit avoids the panic issue by splitting `continuity/testutil/loopback_linux.go`
to another pkg (`continuity/testutil/loopback`), so taht containerd does not need to import
both `containerd/pkg/testutil` and `continuity/testutil`.
i.e. containerd will import `containerd/pkg/testutil` and `continuity/testutil/loopback`.

### DIFF
--- a/fs/dtype_linux_test.go
+++ b/fs/dtype_linux_test.go
@@ -25,6 +25,7 @@ import (
 	"testing"
 
 	"github.com/containerd/continuity/testutil"
+	"github.com/containerd/continuity/testutil/loopback"
 )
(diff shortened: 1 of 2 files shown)


crosbymichael:
LGTM

PR #138, opened 2018-10-29
fs: add WithAllowXAttrErrors CopyOpt

AkihiroSuda:
This option allows ignoring errors during copying xattr like `security.selinux`,
which is not always supported.

Reported in several issues including
* https://github.com/openfaas/openfaas-cloud/issues/312
* https://github.com/moby/buildkit/issues/704
* https://github.com/genuinetools/img/issues/45
* https://bugzilla.redhat.com/show_bug.cgi?id=1596918

### DIFF
--- a/fs/copy.go
+++ b/fs/copy.go
@@ -23,6 +23,7 @@ import (
 	"sync"
 
 	"github.com/pkg/errors"
+	"github.com/sirupsen/logrus"
 )
 
 var bufferPool = &sync.Pool{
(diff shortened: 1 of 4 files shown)


AkihiroSuda:
So the PR appeared to have nothing to do with the buildkit failure but should be still useful

nixpanic:
Please consider [this change](https://github.com/nixpanic/continuity/commit/194c23c791fa3578b7e7117f82a9c0a82b9ccaeb) on top of yours. It continues copying xattrs for a file, even if one of them fails (like `security.selinux`).

Feel free to incorporate the modification in an updated PR. Thanks!

AkihiroSuda:
@nixpanic Thanks, cherry-picked your commit

AkihiroSuda:
@dmcgowan @cpuguy83 PTAL?

cpuguy83:
Oh, nice!
Also related: https://github.com/moby/moby/issues/38155

cpuguy83:
I think for this case, being able to pass in an error handler on file copy would be handy.

Something like:

```go
func(err error) error
```

Where if the handler wants to ignore the error it would return nil.

AkihiroSuda:
Added `type XAttrErrorHandler func(dst, src, xattrKey string, err error) error`. PTAL.

nixpanic:
Hey guys, any progress with reviewing/merging? This PR works well for me and I'd like to see it included so that I can consume it in kubernetes-incubator/external-storage#1013 and a future CSI driver.

Thanks!

fs/copy.go
```diff
@@ -32,14 +33,52 @@ var bufferPool = &sync.Pool{
...
+
+func WithXAttrErrorHandler(xeh XAttrErrorHandler) CopyDirOpt {
+	return func(o *copyDirOpts) error {
+		o.xeh = xeh
+		return nil
+	}
+}
+
+// WithAllowXAttrErrors allows ignoring xattr errors.
+// Ignored errors are logged with logrus.Warnf.
```
mikebrow: is warnf the right log level.. seems this will generate a lot of warnings... maybe debug?

cpuguy83: Agreed.

AkihiroSuda: Maybe s/Warnf/Debugf/g, so that the warnings won't be printed unless debug mode is enabled.

crosbymichael: I usually hate libs that log because there isn't much you can do about it

AkihiroSuda: Removed logrus.

Callers can still provide custom XAttrErrorHandler if logging is needed.

PR #140, opened 2019-04-26
Fix directory comparison in changes

dmcgowan:
Currently string comparison is used to detect changes. The comparisons are already done ordered, however since `-` will compare before `/`, a directory deletion which also has a directory ending with `-` can cause the wrong change output. The comparisons must be done considering directory separators as the lowest sorting value.

Related to https://github.com/moby/buildkit/issues/960

### DIFF
--- a/fs/path_test.go
+++ b/fs/path_test.go
@@ -183,6 +183,51 @@ func TestRootPath(t *testing.T) {
 	t.Run("SymlinkEmpty", testRootPathSymlinkEmpty)
 }
 
+func TestDirectoryCompare(t *testing.T) {
+	for i, tc := range []struct {
+		p1 string
+		p2 string
+		r  int
+	}{
+		{"", "", 0},
+		{"", "/", -1},
+		{"/", "", 1},
+		{"/", "/", 0},
+		{"", "", 0},
+		{"/dir1", "/dir1/", -1},
+		{"/dir1", "/dir1", 0},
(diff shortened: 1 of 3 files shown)


fs/path_test.go
```diff
@@ -183,6 +183,51 @@ func TestRootPath(t *testing.T) {
...
+		{"/dir1/dir2-f1", "/dir1/dir2", 1},
+		{"/dir1/dir2-f1", "/dir1/dir2/", 1},
+		{"/dir1/dir2/", "/dir1/dir2/", 0},
+		{"/dir1/dir2你", "/dir1/dir2a", 1},
+		{"/dir1/dir2你", "/dir1/dir2", 1},
+		{"/dir1/dir2你", "/dir1/dir2/", 1},
+		{"/dir1/dir2你/", "/dir1/dir2/", 1},
+		{"/dir1/dir2你/", "/dir1/dir2你/", 0},
+		{"/dir1/dir2你/", "/dir1/dir2你好/", -1},
+		{"/dir1/dir2你/", "/dir1/dir2你-好/", -1},
```
fuweid: @dmcgowan you should try the character `biang` 😄

AkihiroSuda: no unicode for biang :P

AkihiroSuda:
1 maintainer LGTM and 1 reviewer LGTM. Merging.

PR #141, opened 2019-08-15
Add src string to copyDirectory error message.

WRidder:
Proposal to display the wrongful `src` path in the error message. For example, this error can popup in the Microk8s Kubernetes deployment dashboard. A notion of which src path is referred to would be helpful.


fuweid:
@WRidder thanks for the contribution. we need your sign-off in the commit. `git commit --amend -s --no-edit` might help. :)

fuweid:
ping @estesp @crosbymichael this one is good to go

crosbymichael:
LGTM

PR #142, opened 2019-08-15
Fix fstest missing file updates

dmcgowan:
Updates weren't being considered when checking for directory equality.


crosbymichael:
LGTM

PR #143, opened 2019-08-21
fstest: have CreateSocket actually create a socket

tiborvass:
To create a socket, the Unix listeneter must not be closed,
because that would remove the socket.

Signed-off-by: Tibor Vass <tibor@docker.com>

Ping @ijc @dmcgowan 

FYI I tried to make sure the tests work as intended, but I don't understand how FSSuite works: it doesn't seem to be called from anywhere.


fs/fstest/file.go
```diff
@@ -158,11 +159,15 @@ func Link(oldname, newname string) Applier {
...
 		if err != nil {
 			return err
 		}
-		defer ln.Close()
+		defer unix.Close(fd)
+		sa := &unix.SockaddrUnix{Name: fullPath}
+		if err := unix.Bind(fd, sa); err != nil {
+			return err
+		}
 		return os.Chmod(fullPath, perm)
```
ijc: If this fails you return an error but the socket is left on disk. I suspect this isn't a problem because the caller is using this to populate some sort of temp dir which it will clean up in its entirety on error, but something to be aware of I guess.

tiborvass: I think it's fine, but I'll let @dmcgowan decide

PR #145, opened 2019-11-23
fs: add DiffDirChanges function to get changeset fast

fuweid:
Since AUFS/OverlayFS can persist changeset in diff directory,
DiffDirChanges function can retrieve layer changeset from diff directory
without walking the whole rootfs directory.

Signed-off-by: Wei Fu <fuweid89@gmail.com>

### DIFF
--- a/fs/diff.go
+++ b/fs/diff.go
@@ -18,6 +18,7 @@ package fs
 
 import (
 	"context"
+	"errors"
 	"os"
 	"path/filepath"
 	"strings"
@@ -102,9 +103,6 @@ func Changes(ctx context.Context, a, b string, changeFn ChangeFunc) error {
 	if a == "" {
 		logrus.Debugf("Using single walk diff for %s", b)
 		return addDirChanges(ctx, changeFn, b)
-	} else if diffOptions := detectDirDiff(b, a); diffOptions != nil {
-		logrus.Debugf("Using single walk diff for %s from %s", diffOptions.diffDir, a)
-		return diffDirChanges(ctx, changeFn, a, diffOptions)
 	}
 
 	logrus.Debugf("Using double walk diff for %s from %s", b, a)
@@ -134,24 +132,53 @@ func addDirChanges(ctx context.Context, changeFn ChangeFunc, root string) error 
 	})
 }
 
+// DiffChangeSource is the source of diff directory.
+type DiffSource int
+
+const (
+	// DiffSourceOverlayFS indicates that a diff directory is from
+	// OverlayFS.
+	DiffSourceOverlayFS DiffSource = iota
+)
+
 // diffDirOptions is used when the diff can be directly calculated from
 // a diff directory to its base, without walking both trees.
 type diffDirOptions struct {
-	diffDir      string
-	skipChange   func(string) (bool, error)
-	deleteChange func(string, string, os.FileInfo) (string, error)
+	skipChange   func(string, os.FileInfo) (bool, error)
+	deleteChange func(string, string, os.FileInfo, ChangeFunc) (bool, error)
 }
 
-// diffDirChanges walks the diff directory and compares changes against the base.
-func diffDirChanges(ctx context.Context, changeFn ChangeFunc, base string, o *diffDirOptions) error {
+// DiffDirChanges walks the diff directory and compares changes against the base.
+//
+// NOTE: If all the children of a dir are removed, or that dir are recreated
+// after remove, we will mark non-existing `.wh..opq` file as deleted. It's
+// unlikely to create explicit whiteout files for all the children and all
+// descendants. And based on OCI spec, it's not possible to create a file or
+// dir with a name beginning with `.wh.`. So, after `.wh..opq` file has been
+// deleted, the ChangeFunc, the receiver will add whiteout prefix to create a
+// opaque whiteout `.wh..wh..opq`.
+//
+// REF: https://github.com/opencontainers/image-spec/blob/v1.0/layer.md#whiteouts
+func DiffDirChanges(ctx context.Context, baseDir, diffDir string, source DiffSource, changeFn ChangeFunc) error {
+	var o *diffDirOptions
+
+	switch source {
+	case DiffSourceOverlayFS:
+		o = &diffDirOptions{
+			deleteChange: overlayFSWhiteoutConvert,
+		}
+	default:
+		return errors.New("unknown diff change source")
+	}
+
 	changedDirs := make(map[string]struct{})
-	return filepath.Walk(o.diffDir, func(path string, f os.FileInfo, err error) error {
+	return filepath.Walk(diffDir, func(path string, f os.FileInfo, err error) error {
 		if err != nil {
 			return err
 		}
 
 		// Rebase path
-		path, err = filepath.Rel(o.diffDir, path)
+		path, err = filepath.Rel(diffDir, path)
 		if err != nil {
 			return err
 		}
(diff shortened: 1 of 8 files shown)


fuweid:
The original issue is from https://github.com/moby/buildkit/issues/1192. The docker uses graphdriver to retrieve the layer diff based on AUFS/OverlayFS and it is fast than containerd Compare API which walks the whole rootfs directory. 

The github.com/containerd/continuity/fs package provides the `Changes` function, and it is hard to get relationship between two rootfs paths without original mount information, since the overlayfs mount info might be changed by Chdir if there are too many lowerdirs. Therefore, add new function named by `DiffDirChanges` and containerd Compare API can choose the efficient way to retrieve diff layer.

@dmcgowan @tonistiigi @AkihiroSuda

fs/diff.go
```diff
@@ -134,24 +136,64 @@ func addDirChanges(ctx context.Context, changeFn ChangeFunc, root string) error
...
-func diffDirChanges(ctx context.Context, changeFn ChangeFunc, base string, o *diffDirOptions) error {
+// DiffDirChanges walks the diff directory and compares changes against the base.
+func DiffDirChanges(ctx context.Context, baseDir, diffDir string, source DiffSource, changeFn ChangeFunc) error {
+	var o *diffDirOptions
+
+	switch source {
+	case DiffSourceAUFS:
+		o = &diffDirOptions{
+			skipChange: skipAUFSMetadata,
+		}
```
ktock, line 180: AUFS snapshotter is deprecated since containerd v1.5. Could we focus on overlayfs in this PR?

ktock (commented):
@fuweid Are you still working on this?
Fixing slow diff has been gathering demand in the community (https://github.com/moby/buildkit/issues/1704, https://github.com/moby/buildkit/issues/1192).
Can we move this forward?

fuweid:
> @fuweid Are you still working on this?
> 
> Fixing slow diff has been gathering demand in the community (https://github.com/moby/buildkit/issues/1704, https://github.com/moby/buildkit/issues/1192).
> 
> Can we move this forward?
> 
> 

yeah. will update it in days～ Thanks！

pankajkumar229:
Why was this never merged?

AkihiroSuda:
I think this can be just merged after rebasing and removing AUFS codes

fuweid:
oops. Will update it later. Thanks for remider

fs/fstest/file_windows.go
```diff
@@ -28,6 +29,13 @@ func Lchtimes(name string, atime, mtime time.Time) Applier {
 	})
 }
 
+// CreateDeviceFile provides creates devices Applier.
+func CreateDeviceFile(name string, mode os.FileMode, maj, min int) Applier {
+	return applyFn(func(root string) error {
+		return errors.New("Not implemented")
```
AkihiroSuda, line 35: Nit: should we use `errdefs.ErrNotImplemented`?

fs/diff_test.go
```diff
@@ -350,7 +397,38 @@ func testDiffWithoutBase(t testing.TB, apply fstest.Applier, expected []TestChan
 	return checkChanges(tmp, changes, expected)
 }
 
+func testDiffDirChange(base, diff fstest.Applier, source DiffSource, expected []TestChange) error {
+	baseTmp, err := ioutil.TempDir("", "fast-diff-base-")
```
AkihiroSuda, line 401: ioutil is deprecated

fs/diff_test.go
```diff
@@ -390,6 +468,12 @@ type TestChange struct {
 	Source   string
 }
 
+type TestChanges []TestChange
+
+func (tcs TestChanges) Len() int           { return len(tcs) }
+func (tcs TestChanges) Less(i, j int) bool { return tcs[i].Path < tcs[j].Path }
+func (tcs TestChanges) Swap(i, j int)      { tcs[i], tcs[j] = tcs[j], tcs[i] }
+
```
AkihiroSuda, line 476: Can we just use `sort.Slice`?

fs/diff.go
```diff
@@ -217,19 +253,62 @@ func diffDirChanges(ctx context.Context, changeFn ChangeFunc, base string, o *di
...
+		maj, min, err := devices.DeviceInfo(f)
+		if err != nil {
+			return false, err
+		}
+		return (maj == 0 && min == 0), nil
+	}
+
+	if f.IsDir() {
+		originalPath := filepath.Join(diffDir, path)
+		opaque, err := sysx.LGetxattr(originalPath, "trusted.overlay.opaque")
```
AkihiroSuda, line 291: `user.overlay.*` has to be checked too
https://github.com/torvalds/linux/commit/2d2f2d7322ff43e0fe92bf8cccdc0b09449bf2e1

fs/diff.go
```diff
@@ -217,19 +253,62 @@ func diffDirChanges(ctx context.Context, changeFn ChangeFunc, base string, o *di
...
+		if err != nil {
+			if err == unix.ENODATA {
```
AkihiroSuda, line 293: `errors.Is` is more preferable over directly using the `=` operator

fs/fstest/file_windows.go
fuweid: Good catch. Let me handle it in the follow-up. (Use containerd/errdefs repo to handle this).

fs/diff_test.go
fuweid: Updated

fs/diff_test.go
fuweid: Updated.

fs/diff.go
fuweid: Updated. Please take a look.

fs/diff.go
fuweid: Updated.

AkihiroSuda (approved):

pankajkumar229:
Many Thanks for the quick turnaround.

pankajkumar229:
Is it possible to get this change to a previous version of containerd when it gets released?

AkihiroSuda:
> Is it possible to get this change to a previous version of containerd when it gets released?

Unlikely, due to potential risk of regression.

pankajkumar229:
Is it possible to merge this soon? It will help us compile our own containerd. OTherwise I think we will need some code changes in containerd to use this branch or another fork. 

Thanks again for the quick turnaround.

fs/diff.go
```diff
@@ -163,38 +198,37 @@ func diffDirChanges(ctx context.Context, changeFn ChangeFunc, base string, o *di
...
 
 		var kind ChangeKind
 
-		deletedFile, err := o.deleteChange(o.diffDir, path, f)
-		if err != nil {
-			return err
+		deletedFile := false
+
+		if o.deleteChange != nil {
+			deletedFile, err = o.deleteChange(diffDir, path, f, changeFn)
```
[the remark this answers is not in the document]
fuweid: good catch. Let me think how to handle this

fs/diff_linux.go
```diff
@@ -0,0 +1,97 @@
...
+				return false, nil
+			}
+			return false, err
+		}
+
+		// FIXME(fuweid): In spite of that .wh..wh..opq might not be
+		// existing, mark it as faked delete change so that changeFn
+		// will add whiteout prefix for this faked file. It requires
+		// that changeFn callback must not read the os.FileInfo
+		// parameter.
```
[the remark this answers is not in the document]
fuweid: Good catch. Sorry for late reply. Let me think about how to remove FIXME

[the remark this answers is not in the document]
ktock: If we don't want to introduce a new convention, another approach might be starting a new doubleWalkDiffer on the opaque directory to make changeFn opaque-agnostic.

fs/diff.go
```diff
@@ -102,9 +103,6 @@ func Changes(ctx context.Context, a, b string, changeFn ChangeFunc) error {
 	if a == "" {
 		logrus.Debugf("Using single walk diff for %s", b)
 		return addDirChanges(ctx, changeFn, b)
-	} else if diffOptions := detectDirDiff(b, a); diffOptions != nil {
-		logrus.Debugf("Using single walk diff for %s from %s", diffOptions.diffDir, a)
-		return diffDirChanges(ctx, changeFn, a, diffOptions)
 	}
```
pankajkumar229, line 106: Why did we remove this? It seems we do not call diffDirChanges from fs.Changes anymore. The same delay seems to be there in nerdctl commit.

fuweid: Because you can't detect dir's underlying filesytem based on string path. 
Before comparing two rootfs, we have information about rootfs's mount.
We can check if the filesystem is overlayfs. If so, we can use new function. Otherwise, we should fallback to `fs.Changes`.

pankajkumar229: Thanks

pankajkumar229:
We still see the slowness in commit. It is possibly because "nerdctl commit" is calling the function Compare inside containerd and it seems to create a two temporary mounts: upperRoot and lowerRoot. Both seem to have a full copy of the file system. Now when compare calls writeDiff, even if continuity can check for overlayfs, it is still going through a lot of files and taking forever to do "nerdctl commit". What would be a good way to fix this?
  
Screenshot 2024-01-27 at 3 50 18 PM

Although this belongs to the containerd repository , I thought I would ask here since I saw Mr. Kohei Tokunaga's name in the relevant code and it is possibly better discussed in one place.

pankajkumar229:
I created a (https://github.com/containerd/containerd/pull/9699/files) that essentially does not temp mount both upper and lower but only the lower. And then compares with the upper. It seems to work with the continuity changes in this pull request. Is this the correct approach? It is a pretty small change, just newline changes make it look large.

dmcgowan (approved):

ktock (approved):

PR #149, opened 2020-01-06
syscallx: use golang.org/x/sys

zhsj:
use Readlink from golang.org/x/sys/{unix,windows}

syscall_unix.go is also changed because golang.org/x/sys is
preferred than syscall package.

### DIFF
--- a/syscallx/syscall_unix.go
+++ b/syscallx/syscall_unix.go
@@ -18,9 +18,11 @@
 
 package syscallx
 
-import "syscall"
+import (
+	"golang.org/x/sys/unix"
+)
 
 // Readlink returns the destination of the named symbolic link.
-func Readlink(path string, buf []byte) (n int, err error) {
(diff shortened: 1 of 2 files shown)


estesp:
Given this will no longer use the syscall import, I think it might be better to delete these files/this directory and move the call to sys/unix.Readlink into the `sysx` directory (where it is being used from `file_posix.go`); I guess you would need to create 2 files there given there is windows support, but maybe a readlink_unix.go and a readlink_windows.go in that directory with the calls to sys/unix and sys/windows respectively. I think this directory was temporary while the syscall itself was implemented for Windows before going upstream.

zhsj:
I read the code again, I found the duplicated Readlink on windows is intentional. So let's close this PR.
#113

zhsj:
Although @darstahl  said the fork of Readlink is temporary, but I didn't found upstream commits address any bug. So maybe this is not still not upstream yet.

PR #150, opened 2020-02-27
sysx/xattr: fix and improve

kolyshkin:
### 1. fix `getxattrAll`
    
The current implementation is suboptimal for two reasons:
    
* the initial buffer size of 5 bytes makes no sense
       (my guess @stevvooe forgot to change it back to a sane
       value after the initial retry logic testing);
    
* it never gets the real buffer size, resulting in extra
       iterations (5 -> 10 -> 20 -> ...).
    
This commit
* sets the initial size to `128`;
* uses the logic to get the real size in case we get `ERANGE`
   (rather than doubling the buffer)

### 2. improve `listxattrAll`
The current code
* is hard to read (at least for me, that is);
* is suboptimal (always calls `listxattr()` twice);
* doesn't handle the condition when attributes size
      is changed in between the two calls to `listxattr()`.
    
Fix all three issues at once by reusing the logic from `getxattrAll()`.

### DIFF
--- a/sysx/xattr.go
+++ b/sysx/xattr.go
@@ -20,7 +20,6 @@ package sysx
 
 import (
 	"bytes"
-	"syscall"
 
 	"golang.org/x/sys/unix"
 )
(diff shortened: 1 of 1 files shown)


kolyshkin:
> I like it, but please sign your commits.

@Zyqsempai Sorry, I don't get it. Do you mean something other than the

> Signed-off-by: Kir Kolyshkin <kolyshkin@gmail.com>

line which is present in both commits?

PR #151, opened 2020-02-27
Remove Windows' Readlink fork

kolyshkin:
This reverts the part of commit 8100e750d (PR #113)
that added the fork of `Readlink()` for Windows.

At that time the fork was done to work around the bug
in golang's implementation of `os.Readlink()` for Windows.

The above bug was never reported upstream, but fortunately
it was independently found, reported [1], and fixed [2].
The fix made its way into go-1.13 (there is no mention of
that in release notes, so I checked it in git history).

[1] https://github.com/golang/go/issues/30463
[2] https://go-review.googlesource.com/c/go/+/164201

PS since this requires go 1.13 now, add the appropriate guard
so the package won't compile on earlier golang


kolyshkin:
@crosbymichael @estesp @dmcgowan PTAL

kolyshkin:
@cpuguy83 PTAL

cpuguy83:
I like the change but I'm not sure I like that we are dropping go1.12 support so quickly.

kolyshkin:
> I like the change but I'm not sure I like that we are dropping go1.12 support so quickly.

1. this is windows only
2. go 1.12 is an unsupported release :skull: anyway, since 1.14 is out

so when this change is disruptive, it means someone is using an unsupported golang release

cpuguy83:
Understood that go1.12 is no longer a supported release from the go team, but we still use this in docker 19.03 (for now).

kolyshkin:
So this one is blocked by https://github.com/moby/moby/pull/40592

kolyshkin:
> we still use this in docker 19.03 (for now).

@cpuguy83 we just stopped doing it :) PTAL

crosbymichael:
LGTM

Issue #152, opened 2020-07-01
Empty files can diff as "modified" even when they're not


tomfaulhaber:
This `if` statement is not always triggered correctly because of a flaw in the logic above it.

https://github.com/containerd/continuity/blob/d3ef23f19fbb106bb73ffde425d07a9187e30745/fs/path.go#L126

If the size of the file was 0, `eq` will be false. This means that `sameFile` thinks that two empty files are different even though they are the same.

As a result, when you start a new container and immediately run `ctr snapshot diff` on it, you get a bunch of spurious "changed" empty files. This can be problematic when you're trying to understand if there are real changes in the running image.

cpuguy83:
Fixed in #153 
Thanks!

PR #153, opened 2020-07-01
Fix sameFile() to recognize empty files as the same

tomfaulhaber:
This PR fixes issue #152 and adds a test for that case.

### DIFF
--- a/fs/path.go
+++ b/fs/path.go
@@ -122,6 +122,8 @@ func sameFile(f1, f2 *currentPath) (bool, error) {
 				eq, err = compareSymlinkTarget(f1.fullPath, f2.fullPath)
 			} else if f1.f.Size() > 0 {
 				eq, err = compareFileContent(f1.fullPath, f2.fullPath)
+			} else {
+				eq, err = true, nil
 			}
 			if err != nil || !eq {
 				return eq, err
--- a/fs/path.go
+++ b/fs/path.go
@@ -123,7 +123,7 @@ func sameFile(f1, f2 *currentPath) (bool, error) {
 			} else if f1.f.Size() > 0 {
 				eq, err = compareFileContent(f1.fullPath, f2.fullPath)
 			} else {
-				eq, err = true, nil
+				eq, err = true, nil // if file sizes are zero length, the files are the same by definition
 			}
 			if err != nil || !eq {
 				return eq, err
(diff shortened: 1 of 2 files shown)


fs/path.go
```diff
@@ -122,6 +122,8 @@ func sameFile(f1, f2 *currentPath) (bool, error) {
 				eq, err = compareSymlinkTarget(f1.fullPath, f2.fullPath)
 			} else if f1.f.Size() > 0 {
 				eq, err = compareFileContent(f1.fullPath, f2.fullPath)
+			} else {
+				eq, err = true, nil
```
mikebrow, line 126: nit.. no need to set err=nil.. it's already nil

tomfaulhaber: True. I just did this because it had nice symmetry with the lines above. Happy to change it to `eq = true` if you would prefer.

mikebrow: hmm could start out with eq := true  ..

less else more symmetry

tomfaulhaber: That feels kind of tricky. The else makes it clear that you expect to fall through and you're handling that case. Personally, I'd prefer the else with the `eq = true` to the "implicit fall through." It would make it easier for future readers to see the intent.

mikebrow: if you want less tricky implicit fall throughs 
```
if (f1.f.Mode() & os.ModeSymlink) == os.ModeSymlink {
   eq, err := compareSymlinkTarget(f1.fullPath, f2.fullPath)
   if err != nil || !eq {
      return eq, err
   }
} else if f1.f.Size() > 0 {
   eq, err := compareFileContent(f1.fullPath, f2.fullPath)
   if err != nil || !eq {
      return eq, err
   }
} 
return true, nil // files match || f1.f.Size() == 0
```

mikebrow: even better you could use a switch block

tomfaulhaber: Even in golang, there are a lot of ways to write the same logic. TBH, if were up to me, I'd stick with the first way I wrote it which is a minimal change and very clear to the reader. But I'm happy to go along with whatever the maintainers prefer.

mikebrow: add a comment for your additional "else" case

tomfaulhaber: Good idea. Done.

PR #155, opened 2020-07-07
Migrate to GitHub Actions CI

estesp:
A series of commits to make the initial migration step from Travis CI -> GitHub Actions

- Fix all errors marked by running `golangci-lint` locally with the same settings as other containerd projects
- PRs were not being checked for `gofmt` and `vet` was only run on Linux; move to use all-encompassing `golangci-lint` to replace fmt/vet/lint targets in `Makefile`
- set up the CI workflow for GitHub actions

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,0 +1,131 @@
+name: Continuity
+
+on:
+  push:
+    branches: [ master ]
+  pull_request:
+    branches: [ master ]
+
+jobs:
+
+  checks:
+    name: CI Checks
+    runs-on: ubuntu-18.04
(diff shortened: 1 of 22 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,131 @@
+name: Continuity
+
+on:
+  push:
+    branches: [ master ]
```
mikebrow, line 5: suggest all branches for now or add lint to travis?

estesp: there are no other branches today; if we ever added one we would have to add this action on that branch and configure it, similar to how we handled `release/1.3` adding actions in the main repo.

PR #159, opened 2020-09-17
go.mod: github.com/spf13/cobra v1.0.0

thaJeztah:
Follow-up to https://github.com/containerd/continuity/pull/157 and https://github.com/containerd/continuity/pull/158

keeping separate so that it can be reviewed separately (can close if we don't want this)

PR #161, opened 2020-09-17
go.mod: bazil.org/fuse v0.0.0-20200407214033-5883e5a4b5125

thaJeztah:
rebased to include https://github.com/containerd/continuity/pull/159, as that removed quite a bit of vendor code

This will probably fail miserably on macOS, just noticed this change in the dependency, so opening as a quick draft. We could decide to pin to a more recent version, just before support was removed if macOS support is important, but may be living on borrowed time in case of compatibility issues with other dependencies.

full diff: https://github.com/bazil/fuse/compare/371fbbdaa8987b715bdd21d6adc4c9b20155f748...5883e5a4b512fe2e32f915b1c66a1ddfef81cb3f

This is the last commit before https://github.com/bazil/fuse/commit/60eaf8f021ce00e5c52529cdcba1067e13c1c2c2,
which removed support for macOS (see https://github.com/bazil/fuse/issues/224)

> Remove macOS support
>
> OSXFUSE is no longer open source, see e.g.
> https://colatkinson.site/macos/fuse/2019/09/29/osxfuse/
>
> It looks like Apple is killing the whole category:
> https://news.ycombinator.com/item?id=22251076
>
> I will not personally put uncompensated effort into supporting
> non-open source environments, especially ones that are this hostile to
> work with. The macOS kludges were a significant amount of code, about
> 25% of the remaining non-test lines, and the primary source of
> complexity and alternate code paths.
>
> The API definitions will be remain for a short while, with deprecation
> markers. This should let multiplatform callers work without changes.
> Everything deprecated will be removed soon, so please adjust
> your code.


thaJeztah:
@AkihiroSuda ptal

AkihiroSuda (approved):

dmcgowan (approved):

PR #163, opened 2020-11-19
Fix usage calculation to account for sparse files

dmcgowan:
The current usage calculation does not correctly account for usage within spares files. To correctly account for the usage, the number of blocks must be used rather than the reported file size.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -8,64 +8,23 @@ on:
 
 jobs:
 
-  checks:
-    name: CI Checks
+  #
+  # Project checks
+  #
+  project:
+    name: Project Checks
     runs-on: ubuntu-18.04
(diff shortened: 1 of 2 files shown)


thaJeztah:
@chris-crone you love storage; perhaps you want to review as well?

crosbymichael:
LGTM

chris-crone (commented):
This makes sense to me but I don't know the code well enough to give a proper review :)

justincormack:
I mean, no value is really correct with COW filesystems, tail packing, etc. What is the issue that this is trying to fix (I was told there was one).

dmcgowan:
@justincormack this messes up ephemeral storage accounting in k8s when there are sparse files

mikebrow (approved):
LGTM

PR #164, opened 2020-11-19
gha: replace uses of deprecated "set-env", "add-path"

dmcgowan:
same change as other repos

gha doens't seem to run...


estesp (approved):
LGTM

PR #169, opened 2020-12-04
Fix incorrect usage calculation

dmcgowan:
Set the size based on the correct calculation of blocks and the fixed size unit.
Add tests which align the size with the filesystem block.


tanweernoor (approved):
LGTM

crosbymichael:
LGTM

estesp (approved):
LGTM

PR #171, opened 2020-12-04
Add check to compare against du output

dmcgowan:
Another check which also compares the calculated usage from that reported by calling `du`


crosbymichael:
LGTM

estesp (approved):
LGTM

Issue #172, opened 2020-12-07
TestUsage broken on ZFS


samuelkarp:
When I run `TestUsage` in the `fs` package with a temp dir on a ZFS mount, it fails like this:

```
$ TMPDIR=/testpool/tmp go test ./fs -v -run TestUsage
=== RUN   TestUsage
=== RUN   TestUsage/SingleSmallFile
=== PAUSE TestUsage/SingleSmallFile
=== RUN   TestUsage/MultipleSmallFile
=== PAUSE TestUsage/MultipleSmallFile
=== RUN   TestUsage/BiggerFiles
=== PAUSE TestUsage/BiggerFiles
=== RUN   TestUsage/SparseFiles
=== PAUSE TestUsage/SparseFiles
=== RUN   TestUsage/Hardlinks
=== PAUSE TestUsage/Hardlinks
=== RUN   TestUsage/HardlinkSparefile
=== PAUSE TestUsage/HardlinkSparefile
=== CONT  TestUsage/SingleSmallFile
    du_test.go:129: Wrong usage size 1536, expected 132096
=== CONT  TestUsage/HardlinkSparefile
=== CONT  TestUsage/SparseFiles
    du_test.go:129: Wrong usage size 2560, expected 525312
=== CONT  TestUsage/MultipleSmallFile
=== CONT  TestUsage/Hardlinks
=== CONT  TestUsage/BiggerFiles
=== CONT  TestUsage/HardlinkSparefile
    du_test.go:129: Wrong usage size 1536, expected 263168
=== CONT  TestUsage/MultipleSmallFile
    du_test.go:129: Wrong usage size 2048, expected 263168
=== CONT  TestUsage/Hardlinks
    du_test.go:129: Wrong usage size 1536, expected 132096
=== CONT  TestUsage/BiggerFiles
    du_test.go:129: Wrong usage size 2560, expected 394240
--- FAIL: TestUsage (0.00s)
    --- FAIL: TestUsage/SingleSmallFile (0.00s)
    --- FAIL: TestUsage/SparseFiles (0.00s)
    --- FAIL: TestUsage/Hardlinks (0.00s)
    --- FAIL: TestUsage/BiggerFiles (0.00s)
    --- FAIL: TestUsage/HardlinkSparefile (0.00s)
    --- FAIL: TestUsage/MultipleSmallFile (0.00s)
FAIL
FAIL	github.com/containerd/continuity/fs	0.006s
FAIL
```

[...]

PR #173, opened 2020-12-07
Support FreeBSD

samuelkarp:
This series of commits adds support for FreeBSD.  Aside from `TestUsage` failing on ZFS (see https://github.com/containerd/continuity/issues/172), all the tests are passing.

Some of the files are pretty awkwardly-named; if you have suggestions for changing them I'm happy to do so.

I recognize that there's no CI coverage for FreeBSD today (though adding at least cross-compiling to CI would be reasonably straightforward) and that there may not be interest in maintaining the code.  If that's the case and you prefer not to merge this, I can maintain my own usage of it on my fork.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,6 +76,13 @@ jobs:
       run: make build binaries
       working-directory: src/github.com/containerd/continuity
 
+    - name: Cross-compile
+      if: startsWith(matrix.os, 'ubuntu')
+      shell: bash
+      run: |
+          GOOS=freebsd make build binaries
+      working-directory: src/github.com/containerd/continuity
+
     - name: Linux Tests
       if: startsWith(matrix.os, 'ubuntu')
       run: |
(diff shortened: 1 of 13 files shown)


devices/mknod_linuxdarwinsolaris.go
```diff
@@ -0,0 +1,25 @@
...
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+package devices
+
+import "golang.org/x/sys/unix"
+
+func mknod(path string, mode uint32, dev uint64) (err error) {
+	return unix.Mknod(path, mode, int(dev))
```
mxpv, line 24: `mknod_linuxdarwinsolaris.go` -> should it be just `mknod_unix.go` as default one + `mknod_freebsd.go` as freebsd specific ?

mxpv (commented):
I have a suggestion on file naming, but otherwise LGTM

cpuguy83:
I think making sure it compiles is at least important, and maybe a callout in the readme that freebsd support is untested.

devices/mknod_linuxdarwinsolaris.go
samuelkarp: Done

samuelkarp:
@cpuguy83 

> I think making sure it compiles is at least important

Added to ci.yml

> and maybe a callout in the readme that freebsd support is untested.

I took a stab at wording, but I'm happy to change it if there are any other opinions.

mxpv (approved):

estesp (approved):
LGTM

PR #178, opened 2021-02-26
fs: ability to filter xattr when copying directory

dweomer:
Add fs.WithXAttrExcludes functional option for fs.CopyDir that passes
down a list of xattr keys to filter out when copying.

This ought to allow for a fix for https://github.com/containerd/containerd/issues/5090

Signed-off-by: Jacob Blain Christen <jacob@rancher.com>

### DIFF
--- a/fs/copy.go
+++ b/fs/copy.go
@@ -39,6 +39,8 @@ type XAttrErrorHandler func(dst, src, xattrKey string, err error) error
 
 type copyDirOpts struct {
 	xeh XAttrErrorHandler
+	// xex contains a set of xattrs to exclude when copying
+	xex map[string]struct{}
 }
 
 type CopyDirOpt func(*copyDirOpts) error
@@ -59,6 +61,19 @@ func WithAllowXAttrErrors() CopyDirOpt {
 		return nil
 	}
 	return WithXAttrErrorHandler(xeh)
+}
+
+// WithXAttrExclude allows for exclusion of specified xattr during CopyDir operation.
+func WithXAttrExclude(keys ...string) CopyDirOpt {
+	return func(o *copyDirOpts) error {
+		if o.xex == nil {
+			o.xex = make(map[string]struct{}, len(keys))
+		}
+		for _, key := range keys {
+			o.xex[key] = struct{}{}
+		}
+		return nil
+	}
 }
 
 // CopyDir copies the directory from src to dst.
@@ -104,7 +119,7 @@ func copyDirectory(dst, src string, inodes map[uint64]string, o *copyDirOpts) er
 		return errors.Wrapf(err, "failed to copy file info for %s", dst)
 	}
 
-	if err := copyXAttrs(dst, src, o.xeh); err != nil {
+	if err := copyXAttrs(dst, src, o.xex, o.xeh); err != nil {
 		return errors.Wrap(err, "failed to copy xattrs")
 	}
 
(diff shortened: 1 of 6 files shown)


fs/copy.go
```diff
@@ -39,6 +39,7 @@ type XAttrErrorHandler func(dst, src, xattrKey string, err error) error
 
 type copyDirOpts struct {
 	xeh XAttrErrorHandler
+	xex map[string]bool
```
samuelkarp, line 42: nit: I'd likely have a longer variable name here like `xexclude` or `excludedXattrs` or at least an explanatory comment.  I'd also probably use a `map[string]struct{}` instead of a `map[string]bool`.

```suggestion
	// xex contains a set of xattrs to exclude when copying
	xex map[string]struct{}
```

fs/copy_linux_test.go
```diff
@@ -27,10 +28,88 @@ import (
 	"path/filepath"
 	"testing"
 
+	"github.com/containerd/continuity/sysx"
+
```
samuelkarp, line 32: nit: extra empty line

samuelkarp (approved):
It'd be nice to have some testing on non-Linux Unix since that's also changed here, but I don't think that we have that set up in CI anyway.

A couple small nits, otherwise LGTM.

samuelkarp (approved):
LGTM

fs/copy_linux_test.go
```diff
@@ -27,10 +28,87 @@ import (
...
+		}
+		defer os.RemoveAll(dir)
+		err = CopyDir(dir, src)
+		if err != nil {
+			t.Fatal(err)
+		}
+
+		if one, err := sysx.Getxattr(dir, "user.test-1"); err != nil {
+			t.Fatal(err)
+		} else if string(one) != "one" {
```
cpuguy83, line 66: Can we get rid of the `else if` clasues?

`t.Fatal` will take care of it.

cpuguy83 (changes requested):

fs/copy_linux_test.go
dweomer: Sure. I tend to prefer the limited scope of these kind of if-then-else blocks and the implied lack of redefinition and/or var bleed-through, keeps things tidy. I can work up something a little more mainstream I guess.

dweomer: @cpuguy83 I bit the bullet and adopted the local **fs/fstest** package and moved these changes into fs/copy_unix_test.go as well as implemented a generic-enough-for-our-testing-purposes **assertXAttr** func which uses a switch instead of if-else

AkihiroSuda (approved):

thaJeztah:
@cpuguy83 @dmcgowan ptal

fuweid (approved):
LGTM

estesp (approved):
LGTM

cpuguy83 (approved):
LGTM

fs/copy_unix_test.go
```diff
@@ -0,0 +1,87 @@
+// +build !windows
```
samuelkarp, line 1: `sysx` xattr support is `// +build linux darwin`.  When I attempted to add `freebsd` locally, I got errors like this:

```
copy_unix_test.go:66: failed to copy xattrs: failed to get xattr "\x06test-2\x06test-x\x06test-1" on /tmp/test-copy-src-with-xattr-113390834: attribute not found
```

I think since xattrs are not supported by continuity on FreeBSD, it's better to match the build tags used by the xattr code.  I do realize that I specifically asked you to include FreeBSD earlier; I apologize for asking you to do that without actually knowing that it would work.

```suggestion
// +build linux darwin
```

samuelkarp (changes requested):
One change to build tags that I'd appreciate, otherwise LGTM.

dweomer:
... committing a suggestion breaks DCO?

cpuguy83:
@dweomer Yes, its added as a new commit w/o a sign-off.

AkihiroSuda (approved):

PR #188, opened 2021-11-11
Copy Windows file metadata

gabriel-samfira:
This change ensures that file ownership and Discretionary ACLs are properly preserved on Windows.


fs/copy_linux.go
```diff
@@ -27,7 +27,12 @@ import (
 	"golang.org/x/sys/unix"
 )
 
-func copyFileInfo(fi os.FileInfo, name string) error {
```
dmcgowan, line 30: Why is it necessary to remove `os.FileInfo` from the interface and force an extra stat called? It is not an exported interface, adding the source string is fine if needed for Windows.

gabriel-samfira: Hi Derek,

I was on the fence about this and was hoping for feedback. If you are okay with adding the source string, I'll gladly add it.

gabriel-samfira: Done

jsturtevant:
xref: https://github.com/containerd/containerd/issues/6235

fs/copy_windows.go
```diff
@@ -21,15 +21,56 @@ import (
...
 	}
 
-	// TODO: copy windows specific metadata
+	// Copy file ownership and ACL
+	// We need SeRestorePrivilege and SeTakeOwnershipPrivilege in order
+	// to restore security info on a file, especially if we're trying to
+	// apply security info which includes SIDs not necessarily present on
+	// the host.
+	privileges := []string{winio.SeRestorePrivilege, seTakeOwnershipPrivilege}
+	if err := winio.EnableProcessPrivileges(privileges); err != nil {
```
jsturtevant, line 44: Something similar is done during snapshotting:
 https://github.com/containerd/containerd/blob/7758cdc09ad5c6c5e5eecdfacfa639d3cf87bdbf/snapshots/windows/windows.go#L474-L480

kzys (approved):

dmcgowan (approved):

PR #190, opened 2021-12-03
Build containerd/continuity on multiple Unix OSes

### LINKED ISSUE #183: Support GOOS=netbsd
The `fs` package is not compilable for `GOOS=netbsd`, as of the current main branch (7f427325e878a1890b681ae71c7a43b1ef076a5f) .

```console
$ GOOS=netbsd go build ./fs/
# github.com/containerd/continuity/fs
fs/copy.go:118:12: undefined: copyFileInfo
fs/copy.go:122:12: undefined: copyXAttrs
fs/copy.go:157:14: undefined: copyDevice
fs/copy.go:164:13: undefined: copyFileInfo
fs/copy.go:168:13: undefined: copyXAttrs
fs/copy.go:190:9: undefined: copyFileContent
fs/diff_unix.go:58:26: undefined: sysx.ENODATA
fs/diff_unix.go:62:26: undefined: sysx.ENODATA
```

kzys:
While the .go files have multiple build tags, they haven't been tested on
GitHub Actions except for FreeBSD.

This change also fixes #183.

Signed-off-by: Kazuyoshi Kato <katokazu@amazon.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -76,13 +76,6 @@ jobs:
       run: make build binaries
       working-directory: src/github.com/containerd/continuity
 
-    - name: Cross-compile
-      if: startsWith(matrix.os, 'ubuntu')
-      shell: bash
-      run: |
-          GOOS=freebsd make build binaries
-      working-directory: src/github.com/containerd/continuity
-
     - name: Linux Tests
       if: startsWith(matrix.os, 'ubuntu')
       run: |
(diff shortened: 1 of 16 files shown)


dmcgowan (approved):

samuelkarp:
Other than just cross-compiling, have we done any testing on those other operating systems?  (In other words: should we include build tags and thus "support" for an operating system that is not tested?)

kzys:
No. We technically could use GitHub Actions' macOS builder to run VirtualBox (like we do in containerd), but I haven't tried that yet.

I still think this is better than what we have today. We have a bunch of build tags but the combinations are not really tested.

samuelkarp:
I was going to say that it looks like openbsd and netbsd are new additions and maybe we should let compilation fail, but I see in https://github.com/containerd/continuity/issues/183 that @AkihiroSuda is trying to use continuity on netbsd already.

Looks like there's already a note in the readme about support, so I think this is fine.

samuelkarp (approved):

kzys:
@estesp Can you take a look? This should be okay to merge LGTM-wise.

estesp (approved):
LGTM

PR #191, opened 2021-12-03
Do not log errors before returning them

### LINKED ISSUE #179: context.Resource generates unavoidable log-spam on Windows
Noted in https://github.com/containerd/containerd/pull/4419 where this is generating 140k lines of log output for a _passing_ test suite.

`func (c *context) Resource(p string, fi os.FileInfo) (Resource, error)` and its subordinate `func (c *context) resolveXAttrs(fp string, fi os.FileInfo, base *resource) (map[string][]byte, error)` contain `log.Printf` or `log.Println` in non-error paths, which lead to unavoidable log-spam when using them on a system like Windows, which will see two log outputs for every file touched, even though [one of these code paths is not an error](https://github.com/containerd/continuity/blob/f2cc35102c2a086e89ea40de1c0a99861713c51b/context.go#L155-L159).

It would be better for most of those to use [`github.com/pkg/errors.Wrap` ](https://pkg.go.dev/github.com/pkg/errors#hdr-Adding_context_to_an_error) or [`fmt.Errorf`'s `%w` verb](https://blog.golang.org/go1.13-errors) to capture the desired error, and if ignoring it, ignore it silently. `github.com/pkg/errors` is already in _go.mod_, so this would not be a new dependency.

As a fallback, perhaps being able to determine earlier (by testing or parameter) that xattr support is not present, logging it _once_ per context rather than once per `Resource` might help.

I note a similiar issue in [`func BuildManifest(ctx Context) (*Manifest, error)`](https://github.com/containerd/continuity/blob/f2cc35102c2a086e89ea40de1c0a99861713c51b/manifest.go#L95). A few lines later is a `fmt.Errorf` using `%v` that could use `%w`, exposing the underlying error programmatically.

kzys:
Except for one case, all logs are ultimately returned to the callers.
Fixes #179.

Signed-off-by: Kazuyoshi Kato <katokazu@amazon.com>

### DIFF
--- a/context.go
+++ b/context.go
@@ -20,7 +20,6 @@ import (
 	"bytes"
 	"fmt"
 	"io"
-	"log"
 	"os"
 	"path/filepath"
 	"strings"
@@ -152,9 +151,7 @@ func (c *context) Resource(p string, fi os.FileInfo) (Resource, error) {
 	}
 
 	base.xattrs, err = c.resolveXAttrs(fp, fi, base)
-	if err == ErrNotSupported {
-		log.Printf("resolving xattrs on %s not supported", fp)
-	} else if err != nil {
+	if err != nil && err != ErrNotSupported {
 		return nil, err
 	}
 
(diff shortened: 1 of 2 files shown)


kzys:
cc @TBBle

TBBle (approved):

dmcgowan (approved):

PR #212, opened 2023-02-04
Add paths to windows metadataFiles

gabriel-samfira:
When creating a new empty base layer on Windows, hcsshim creates a set of empty registry hives in C:\Windows\System32\config. These files are needed in order for the new layer to be considered valid.

Add these locations to the metadataFiles, which excludes them when comparing the diffs of two layers.

This will be used in the snapshot tests on Windows in: https://github.com/containerd/containerd/pull/8043 

Signed-off-by: Gabriel Adrian Samfira <gsamfira@cloudbasesolutions.com>


gabriel-samfira:
These files get created here: https://github.com/microsoft/hcsshim/pull/1637/files#diff-94af4df272f2d4cf11cff84cc70fe9ac3a715b0e2c13440bb5e131f56660806cR23

dcantah (approved):

fs/fstest/compare_windows.go
```diff
@@ -19,6 +19,14 @@ package fstest
 // TODO: Any more metadata files generated by Windows layers?
```
estesp, line 19: nit: should this TODO: line be removed now that the additional files are added by this PR?

dcantah: Could argue there still may be more 😅 but I agree, probably time to get rid of this.

gabriel-samfira: indeed. There may be more, but we can just add them as we discover them. Removing the TODOs

gabriel-samfira: done

estesp (approved):
LGTM

austinvazquez (approved):

kzys (approved):

PR #216, opened 2023-02-24
Update GitHub Actions runner images

austinvazquez:
Resolves deprecation warnings in CI by moving macOS from 10.15 to 12.

Signed-off-by: Austin Vazquez <macedonv@amazon.com>

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
 
     strategy:
       matrix:
-        os: [ubuntu-22.04, macos-10.15, windows-2019]
+        os: [ubuntu-22.04, macos-12, windows-2019, windows-2022]
 
     steps:
     - name: Set up Go


samuelkarp (approved):

estesp (approved):

estesp:
Does anyone from the Windows side have an opinion on whether it's better to test both windows-2019 and windows-2022 or OK to move to windows-2022 only? @kevpar?

kevpar:
> Does anyone from the Windows side have an opinion on whether it's better to test both windows-2019 and windows-2022 or OK to move to windows-2022 only? @kevpar?

I'd recommend we keep both windows-2019 and windows-2022 for the time being. There are a lot of container users still on 2019.

austinvazquez:
> > Does anyone from the Windows side have an opinion on whether it's better to test both windows-2019 and windows-2022 or OK to move to windows-2022 only? @kevpar?
> 
> I'd recommend we keep both windows-2019 and windows-2022 for the time being. There are a lot of container users still on 2019.

Sounds good. Thanks @kevpar. Will revert and add the new Windows runner to the testing matrix.

estesp (approved):

PR #220, opened 2023-04-20
Enable tests for all platforms

dmcgowan:
Try to run tests on more platforms

### DIFF
--- a/fs/copy_unix.go
+++ b/fs/copy_unix.go
@@ -23,6 +23,7 @@ import (
 	"fmt"
 	"io"
 	"os"
+	"runtime"
 	"syscall"
 
 	"github.com/containerd/continuity/sysx"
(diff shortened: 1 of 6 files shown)


fs/copy_unix.go
```diff
@@ -71,6 +71,10 @@ func copyFileContent(dst, src *os.File) error {
 func copyXAttrs(dst, src string, excludes map[string]struct{}, errorHandler XAttrErrorHandler) error {
 	xattrKeys, err := sysx.LListxattr(src)
 	if err != nil {
+		if os.IsPermission(err) {
```
samuelkarp, line 74:
```suggestion
		if os.IsPermission(err) && runtime.GOOS == "darwin" {
```
?

samuelkarp (commented):
I've been meaning to do FreeBSD tests for this too on Cirrus, though that can be a separate PR.

samuelkarp:
It'd also be good to test more filesystems: https://github.com/containerd/continuity/issues/172

dmcgowan:
@samuelkarp agreed, would be a little bit more involved. I'm not completely sure about the Windows test, at least what has changed, we should at least have the tests pass for our CI environment. Expanding to more environments would be good though given how the fs library is used.

samuelkarp (approved):

samuelkarp:
I opened https://github.com/containerd/continuity/issues/221 to track adding coverage for more filesystems.

estesp (approved):

PR #227, opened 2023-06-09
fs: use io.Copy because go supports CopyFileRange

fuweid:
REF: https://github.com/golang/go/commit/7be3f09deb2dc1d57cfc18b18e12192be3544794

### DIFF
--- a/fs/copy.go
+++ b/fs/copy.go
@@ -18,19 +18,12 @@ package fs
 
 import (
 	"fmt"
+	"io"
 	"os"
 	"path/filepath"
-	"sync"
 
 	"github.com/sirupsen/logrus"
 )
-
-var bufferPool = &sync.Pool{
-	New: func() interface{} {
-		buffer := make([]byte, 32*1024)
-		return &buffer
-	},
-}
 
 // XAttrErrorHandler transform a non-nil xattr error.
 // Return nil to ignore an error.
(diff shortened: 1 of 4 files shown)


dmcgowan:
I wonder if we could just get rid of this function all together, for unix we still do 
```
func copyFileContent(dst, src *os.File) error {
	buf := bufferPool.Get().(*[]byte)
	_, err := io.CopyBuffer(dst, src, *buf)
	bufferPool.Put(buf)

	return err
}
```

Based on the Go code though, this buffer will never be used for os.File to os.File copying though

fuweid:
> I wonder if we could just get rid of this function all together

If the go runtime move the `ReadFrom` into linux platform, I guess it will be possible

fs/copy.go
```diff
@@ -199,5 +201,22 @@ func openAndCopyFile(target, source string) error {
 	}
 	defer tgt.Close()
 
-	return copyFileContent(tgt, src)
+	var tgtWriter io.Writer = tgt
+	if runtime.GOOS != "linux" {
+		tgtWriter = onlyWriter{tgt}
+	}
+
+	buf := bufferPool.Get().(*[]byte)
```
dmcgowan, line 209: The buffer didn't even end up getting used before, should we even bother with it rather than checking OS. It could be possible for Go to implement copy optimizations on other platforms as well.

fuweid: Updated. Let's use `io.Copy` directly.

kzys (approved):

PR #242, opened 2024-06-22
gha: run CI on go1.22

thaJeztah:

PR #245, opened 2024-07-16
fs: properly handle ENOTSUP in copyXAttrs

### LINKED ISSUE #244: [Bug] CopyDir fails on systems without xattr support
The [CopyDir()](https://github.com/containerd/continuity/blob/50fa7de4fc5d1529fed1c4d6e3efad231bf5a232/fs/copy.go#L74) call will fail on filesystems that do not support extended attributes. This is because the function will eventually call [sysx.LListxattr()](https://github.com/containerd/continuity/blob/50fa7de4fc5d1529fed1c4d6e3efad231bf5a232/fs/copy_linux.go#L65), which will return ENOTSUP. This shouldn't be a problem for normal operations, as it just means there are no xattrs to copy, but as it's still an error it will say the function call failed.

This can be mitigated by attaching an XAttrErrorHandler to the CopyDir function call, but this being default behavior seems to be a bug.

sondavidb:
Fixes #244. More context in the issue.

Filesystems without xattr support do not have any xattrs to copy. The syscall for this will return ENOTSUP to indicate this, but continuity will treat it as a regular error. This change will return nil if this call returns ENOTSUP.

### DIFF
--- a/fs/copy_linux.go
+++ b/fs/copy_linux.go
@@ -17,6 +17,7 @@
 package fs
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"syscall"
@@ -64,6 +65,9 @@ func copyFileInfo(fi os.FileInfo, src, name string) error {
 func copyXAttrs(dst, src string, excludes map[string]struct{}, errorHandler XAttrErrorHandler) error {
 	xattrKeys, err := sysx.LListxattr(src)
 	if err != nil {
+		if errors.Is(err, unix.ENOTSUP) {
+			return nil
+		}
 		e := fmt.Errorf("failed to list xattrs on %s: %w", src, err)
 		if errorHandler != nil {
 			e = errorHandler(dst, src, "", e)
--- a/fs/copy_unix.go
+++ b/fs/copy_unix.go
@@ -20,12 +20,14 @@
 package fs
 
 import (
+	"errors"
 	"fmt"
 	"os"
 	"runtime"
 	"syscall"
 
 	"github.com/containerd/continuity/sysx"
+	"golang.org/x/sys/unix"
 )
 
 func copyFileInfo(fi os.FileInfo, src, name string) error {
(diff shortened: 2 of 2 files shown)


estesp (approved):
LGTM

thaJeztah (approved):
LGTM

fs/copy_linux.go
```diff
@@ -64,6 +64,9 @@ func copyFileInfo(fi os.FileInfo, src, name string) error {
 func copyXAttrs(dst, src string, excludes map[string]struct{}, errorHandler XAttrErrorHandler) error {
 	xattrKeys, err := sysx.LListxattr(src)
 	if err != nil {
+		if err == unix.ENOTSUP {
```
thaJeztah, line 67: Oh! Actually; this is handling an error returned by `sysx`,  which is not a raw stdlib / syscall package, perhaps better to use `errors.Is()` here?

Some discussion with @robmry and @djs55 (we happened to be looking into a similar issue), and there's some discussion if `EOPNOTSUPP` should be handled here; this was related to https://elixir.bootlin.com/linux/v6.10/source/fs/fuse/xattr.c#L23

sondavidb: > perhaps better to use errors.Is() here

Makes sense, pushed the change

> there's some discussion if `EOPNOTSUPP` should be handled here

It seems pretty reasonable to me, since the `fuse_getxattr` call would also fail in this use case. LMK and I can add add a check for this condition

estesp:
haven't had time to look deeper, but this CI test failure is strange; I don't think we have flakes in this project, so is this being caused by the change:
```
=== RUN   TestDiffDirChangeWithOverlayfs
    diff_test.go:249: failed diff dir change: Unexpected number of changes:
        got(5):
        	modify	/dir1
        	delete	/dir1/d
        	modify	/dir1/f
        	modify	/dir2/d/f
        	delete	/dir3/.wh..opq
        expected(8):
        	modify	/dir1
        	delete	/dir1/d
        	modify	/dir1/f
        	modify	/dir2
        	modify	/dir2/d
        	modify	/dir2/d/f
        	modify	/dir3
        	delete	/dir3/.wh..opq
--- FAIL: TestDiffDirChangeWithOverlayfs (0.00s)
```

thaJeztah:
@estesp I think I saw the same failure on https://github.com/containerd/continuity/pull/242

wondering if it's changes on the runners

AkihiroSuda:
Please rebase to pass the CI

Issue #247, opened 2024-10-20
`TestDiffDirChangeWithOverlayfs` is failing (when `TMPDIR` is on tmpfs, not on ext4)


AkihiroSuda:
> Interesting; looks like this is failing on go1.22 
> 
> cc @AkihiroSuda @dmcgowan 
> 
> ```
> === RUN   TestDiffDirChangeWithOverlayfs
>     diff_test.go:249: failed diff dir change: Unexpected number of changes:
>         got(6):
>         	modify	/dir1
>         	delete	/dir1/d
>         	modify	/dir1/f
>         	modify	/dir2/d/f
>         	modify	/dir3
>         	delete	/dir3/.wh..opq
>         expected(8):
>         	modify	/dir1
>         	delete	/dir1/d
>         	modify	/dir1/f
>         	modify	/dir2
>         	modify	/dir2/d
>         	modify	/dir2/d/f
>         	modify	/dir3
>         	delete	/dir3/.wh..opq
> --- FAIL: TestDiffDirChangeWithOverlayfs (0.01s)
> 

_Originally posted by @thaJeztah in https://github.com/containerd/continuity/issues/242#issuecomment-2183597523_
            
(Unlikely specific to Go 1.22)

AkihiroSuda:
Locally reproducible on Ubuntu 24.10 (kernel `6.11.0-9-generic #9-Ubuntu`), but not reproducible on Ubuntu 24.04 (kernel `6.8.0-45-generic #45-Ubuntu`).

Not sure if this difference arises from the kernel version or some filesystem configuration.

AkihiroSuda:
Looks like the test depends on the filesystem of `TMPDIR`, not on the kernel version.

Works with:
- `/dev/vda1 on / type ext4 (rw,relatime,discard,errors=remount-ro,commit=30)`

Fails with:
- `tmpfs on /tmp type tmpfs (rw,nosuid,nodev,size=2003352k,nr_inodes=1048576,inode64)`

AkihiroSuda:
PR https://github.com/containerd/continuity/pull/249

PR #248, opened 2024-10-20
Investigate TestDiffDirChangeWithOverlayfs failure

AkihiroSuda:
#247


AkihiroSuda:
The test fails on Ubuntu 20.04 too.
Probably something was changed on the kernel side and it was backported to Ubuntu 20.04?
cc @thaJeztah @dmcgowan

PR #280, opened 2026-04-12
Add native Go ext4 filesystem image creator

dmcgowan:
Pure Go implementation of ext4 formatting that produces valid, optimized sparse images without requiring mkfs.ext4 or any host binaries. Output is byte-comparable to mkfs.ext4.

This is designed to be used with the erofs snapshotter to remove the dependency on a host binary and speed up upper creation to speed up overall container startup time.

### Why not an existing implementation?

The only Go library that supports ext4 creation is [go-diskfs](https://github.com/diskfs/go-diskfs). Other Go ext4 packages ([dsoprea/go-ext4](https://github.com/dsoprea/go-ext4), [masahiro331/go-ext4-filesystem](https://github.com/masahiro331/go-ext4-filesystem)) are read-only. Microsoft's [hcsshim/ext4/tar2ext4](https://github.com/microsoft/hcsshim) converts tarballs to ext4 but does not support creating empty formatted images.

go-diskfs is a general-purpose disk and filesystem library (~9,000 lines for ext4 alone, plus FAT12/16/32, ISO 9660, squashfs, and GPT/MBR partitioning). It seems to be designed for writing to block devices, not creating sparse file images. Several aspects make it impractical for our use case:

**Eager zero-fill architecture.** go-diskfs explicitly writes zeros to every inode table for every block group (`ext4.go:3069`), every bitmap (`ext4.go:3045-3057`), and the entire journal in 1 MiB chunks (`ext4.go:2800-2816`). This destroys file sparseness — a 256 MiB image consumes 12 MiB on disk vs 4 MiB with sparse writes. Changing this requires rewriting the creation path across `initGroupDescriptorTables`, `buildGroupDescriptorsFromSuperblock`, `initJournal`, and the `Create` orchestration (~300 lines spanning multiple functions). The layout calculations could survive, but the write path would be entirely new.

**Fails e2fsck at 256 MiB and above.** Empty block groups are never marked with `INODE_UNINIT` or `BLOCK_UNINIT` flags (`ext4.go:3367`). With default settings, creation crashes at 1 GiB with a journal extent tree error. This indicates the library has not been heavily exercised at sizes relevant to container images.

**18 transitive dependencies.** Four compression libraries (lz4, xz, lzo, zstd), logrus, xattr handling, and others — none used by the ext4 code path.

### Benchmarks (256 MiB image)

| | Native `Format` | `mkfs.ext4` (optimized) | `mkfs.ext4` (defaults) | go-diskfs |
|---|---|---|---|---|
| **Time** | 1.3 ms | 5.0 ms | 9.5 ms | 26 ms |
| **Disk usage** | 4 MiB | 6 MiB | 12 MiB | 12 MiB |
| **e2fsck** | pass | pass | pass | fail |
| **Host binaries** | none | mkfs.ext4 | mkfs.ext4 | none |
| **Dependencies** | stdlib only | - | - | 18 modules |

The native formatter performs the same work as `mkfs.ext4` — the speed difference vs `mkfs.ext4` (optimized) is primarily fork/exec and process startup overhead, not the formatting itself. The `mkfs.ext4` C implementation does the same sparse-aware writes under the same optimized flags. The native formatter avoids this process overhead entirely and requires no host binaries, while producing byte-comparable output (identical superblock fields, block layout, and bitmap content — differing only in UUID and derived checksums). go-diskfs is 20x slower due to its eager zero-fill write path.

### Features

- Pure Go, no external binaries required
- Sparse file output — only non-zero metadata blocks written to disk
- CRC32C metadata checksums (superblock, GDT, bitmaps, inodes, directory blocks)
- Multi-group support with flex_bg metadata consolidation
- `INODE_UNINIT` / `BLOCK_UNINIT` flags on empty groups
- Optional directory creation with permissions and ownership
- Parent directories created automatically, inheriting the child's permissions

### DIFF
--- a/pkg/ext4/format.go
+++ b/pkg/ext4/format.go
@@ -1,0 +1,1125 @@
+/*
+   Copyright The containerd Authors.
+
+   Licensed under the Apache License, Version 2.0 (the "License");
+   you may not use this file except in compliance with the License.
+   You may obtain a copy of the License at
+
+       http://www.apache.org/licenses/LICENSE-2.0
+
+   Unless required by applicable law or agreed to in writing, software
+   distributed under the License is distributed on an "AS IS" BASIS,
+   WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+   See the License for the specific language governing permissions and
+   limitations under the License.
+*/
+
+// Package ext4 creates ext4 filesystem images.
+package ext4
+
+import (
+	"crypto/rand"
+	"encoding/binary"
+	"fmt"
+	"hash/crc32"
+	"io"
+	"io/fs"
+	"os"
+	"strings"
+	"time"
+)
+
+// CreateOpt configures filesystem creation.
+type CreateOpt func(*createConfig)
+
+type createConfig struct {
+	dirs []dirEntry
+}
+
+// dirEntry represents a directory to be created in the filesystem
+// with specific permissions and ownership.
+type dirEntry struct {
+	Path string
+	Mode fs.FileMode
+	UID  uint32
+	GID  uint32
+}
+
+// WithDir adds a directory to be created in the filesystem with the
+// specified permissions and ownership. Parent directories are created
+// automatically, inheriting the child's permissions.
+func WithDir(path string, mode fs.FileMode, uid, gid uint32) CreateOpt {
+	return func(c *createConfig) {
+		c.dirs = append(c.dirs, dirEntry{Path: path, Mode: mode, UID: uid, GID: gid})
+	}
+}
+
+// splitPath splits a path into its components, ignoring leading/trailing slashes.
+func splitPath(p string) []string {
+	p = strings.Trim(p, "/")
+	if p == "" {
+		return nil
+	}
+	return strings.Split(p, "/")
+}
+
+const (
+	superblockOffset = 1024
+	superblockSize   = 1024
+	ext4Magic        = 0xEF53
+
+	blockSize       = 4096
+	inodeSize       = 256
+	groupDescSize   = 64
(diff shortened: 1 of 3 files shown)


pkg/ext4/format.go
```diff
@@ -0,0 +1,1131 @@
...
+}
+
+// dirEntryRecLen returns the minimum record length for a dir entry with the given name.
+// Must be a multiple of 4.
+func dirEntryRecLen(name string) uint16 {
+	// 8 bytes header + name length, rounded up to 4
+	return uint16((8 + len(name) + 3) &^ 3)
+}
+
+func writeLostFound(f *os.File, p *formatParams, sb *superblock) error {
```
hsiangkao: it can be called as `writeEmptyDir`

hsiangkao (approved):
I like this idea

pkg/ext4/format.go
```diff
@@ -0,0 +1,1131 @@
...
+		startByte++
+	}
+	for i := startByte; i < blockSize; i++ {
+		buf[i] = 0xFF
+	}
+
+	return buf
+}
+
+func writeRootDir(f *os.File, p *formatParams, sb *superblock) error {
```
hsiangkao, line 855: `writeRootDir` and `writeUserDirs` can be merged.
We can move `for _, d := range p.userDirs {` into the caller.

dmcgowan: I removed the loop but there is enough differences in `writeRootDir` and `writeUserDirs` that merging would make the function more complicated.

hsiangkao: ok, if it becomes complicated, let's leave it as-is.

austinvazquez (approved):

PR #281, opened 2026-04-12
Remove fstest dependency on root package

dmcgowan:
Related to #277

## Summary

- Replace `fstest`'s use of `continuity.NewContext`, `continuity.BuildManifest`, and the `continuity.Resource` interfaces with a local directory walker and plain resource struct
- Break the transitive dependency chain where consumers of `CheckDirectoryEqual` (used in ~37 test files across containerd and buildkit) pull in the full root package with its protobuf, manifest, and driver machinery
- Add test coverage for `CheckDirectoryEqual`, `CheckDirectoryEqualWithApplier`, and the new `buildResources` walker

## Details

The `fs/fstest` package provides `CheckDirectoryEqual` for comparing two directory trees in tests. Previously this worked by building full continuity manifests (context + manifest + protobuf resources), which dragged in the entire root `continuity` package as a dependency. Any external repo importing `fs/fstest` for test helpers transitively required protobuf, the driver framework, digest libraries, and other machinery unrelated to simple directory comparison.

The new implementation replaces this with:
- `walker.go` — a `filepath.Walk`-based directory walker that produces a sorted `[]resource` list with path, mode, uid/gid, size, SHA256 content hash, symlink target, and device major/minor
- `walker_unix.go` / `walker_windows.go` — platform-specific stat, hardlink detection, and device info extraction
- `continuity_util.go` — updated diff and comparison functions operating on the concrete `resource` type instead of `continuity.Resource` interfaces

The comparison behavior is preserved:
- Path, mode, uid, gid are compared for all resource types
- Regular files compare size, SHA256 digest, and hardlink paths
- Symlinks compare targets
- Devices compare major/minor numbers
- Directories and named pipes compare by metadata only
- XAttrs were collected but never compared before (`// TODO` in old code) and remain uncompared

## Test plan

- [x] New tests in `compare_test.go` cover: identical directories, content differences, extra/missing files, symlinks, symlink target differences, hardlinks, permission differences, applier-based comparison, and `buildResources` output validation
- [x] `go test ./fs/fstest/` passes
- [x] Existing `fs/` test suite (`TestDirectoryCompare`, copy tests using `CheckDirectoryEqual`) continues to pass

### DIFF
--- a/fs/fstest/continuity_util.go
+++ b/fs/fstest/continuity_util.go
@@ -19,25 +19,24 @@ package fstest
 import (
 	"bytes"
 	"fmt"
-
-	"github.com/containerd/continuity"
+	"os"
 )
 
 type resourceUpdate struct {
-	Original continuity.Resource
-	Updated  continuity.Resource
+	Original resource
+	Updated  resource
 }
 
 func (u resourceUpdate) String() string {
 	return fmt.Sprintf("%s(mode: %o, uid: %d, gid: %d) -> %s(mode: %o, uid: %d, gid: %d)",
-		u.Original.Path(), u.Original.Mode(), u.Original.UID(), u.Original.GID(),
-		u.Updated.Path(), u.Updated.Mode(), u.Updated.UID(), u.Updated.GID(),
+		u.Original.path, u.Original.mode, u.Original.uid, u.Original.gid,
+		u.Updated.path, u.Updated.mode, u.Updated.uid, u.Updated.gid,
 	)
 }
 
 type resourceListDifference struct {
-	Additions []continuity.Resource
-	Deletions []continuity.Resource
+	Additions []resource
+	Deletions []resource
 	Updates   []resourceUpdate
 }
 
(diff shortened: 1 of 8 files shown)


fs/fstest/continuity_util.go
```diff
@@ -112,103 +111,51 @@ func diffResourceList(r1, r2 []continuity.Resource) resourceListDifference {
...
 }
 
-func compareResourceTypes(r1, r2 continuity.Resource) bool {
-	switch t1 := r1.(type) {
-	case continuity.RegularFile:
-		t2, ok := r2.(continuity.RegularFile)
-		if !ok {
+func compareResourceType(r1, r2 resource) bool {
+	mode := r1.mode
+	switch {
```
djdongjin, line 133: nit
```suggestion
	switch mode := r1.mode {
```

djdongjin (approved):

AkihiroSuda (approved):

PR #292, opened 2026-06-07
fs: defensive guard invalid FileInfo stat in DiskUsage

dyer-oai:
## Summary

Guard `DiskUsage` and `DiffUsage` against invalid `os.FileInfo.Sys()` values before dereferencing `*syscall.Stat_t`.

## Root cause

`fs.diskUsage` and `fs.diffUsage` unconditionally asserted and dereferenced `fi.Sys().(*syscall.Stat_t)`. A typed-nil or otherwise invalid `Sys()` payload therefore panicked instead of returning an error.

I hit this through containerd `1.7.27` while CRI snapshot stats were walking an active overlay snapshot:

```text
github.com/containerd/continuity/fs.diskUsage.func1(...)
.../vendor/github.com/containerd/continuity/fs/du_unix.go:71
github.com/containerd/containerd/snapshots/overlay.(*snapshotter).Usage(...)
github.com/containerd/containerd/pkg/cri/server.(*snapshotsSyncer).sync(...)
```

Because that happened inside containerd's background snapshot stats sync, the panic took down the containerd daemon instead of being reported as a snapshot usage error.

## Validation

- `go test ./fs`
- `go test ./...`

### DIFF
--- a/fs/du_unix.go
+++ b/fs/du_unix.go
@@ -20,6 +20,7 @@ package fs
 
 import (
 	"context"
+	"fmt"
 	"os"
 	"path/filepath"
 	"syscall"
@@ -47,6 +48,16 @@ func newInode(stat *syscall.Stat_t) inode {
 	}
 }
 
+func fileInfoStat(path string, fi os.FileInfo) (*syscall.Stat_t, error) {
+	sys := fi.Sys()
+	stat, ok := sys.(*syscall.Stat_t)
+	if !ok || stat == nil {
+		return nil, fmt.Errorf("failed to get stat for %q: expected *syscall.Stat_t, got %T", path, sys)
+	}
+
+	return stat, nil
+}
+
 func diskUsage(ctx context.Context, roots ...string) (Usage, error) {
 	var (
 		size   int64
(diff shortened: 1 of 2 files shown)


dyer-oai:
This came from a real containerd 1.7.27 panic in CRI snapshot stats while walking an active overlay snapshot:

```text
github.com/containerd/continuity/fs.diskUsage.func1
.../du_unix.go:71
github.com/containerd/containerd/snapshots/overlay.(*snapshotter).Usage
github.com/containerd/containerd/pkg/cri/server.(*snapshotsSyncer).sync
```

I could not reproduce a standing bad filesystem condition on the node after reboot; the current ext4/overlayfs snapshot tree walks with `bad_stat=0`. This patch is meant to turn that impossible/rare `FileInfo.Sys()` case into an error instead of killing containerd.

dyer-oai:
@dmcgowan @djdongjin would appreciate a review when you get a chance. This is the small defensive fix for the containerd 1.7.27 `DiskUsage` panic described above.

djdongjin (approved):
LGTM

dmcgowan (approved):

PR #310, opened 2026-07-23
fs: make fsync opt-in via CopyFileOpt and CopyDirOpt

### LINKED ISSUE #303: fs: per-file fsync in sparse-aware copyFile causes 17x regression for many-files workloads
### Problem

Commit a424ba1 ("Linux: Make copyFile sparse aware") introduced `tgt.Sync()` at the end of the new `copyFile` implementation in `fs/copy_linux.go`. While the sparse-aware copy logic itself is valuable, the per-file fsync causes a severe performance regression when copying directories with many files.

In my use case, this caused significant container start latency when using Docker-style volumes (`VOLUME` directive in Dockerfile). The container runtime's `volumeCopyUp` path uses `fs.CopyDir` to copy image content into the volume mount at container start. When the image has thousands of files in the volume path, the per-file fsync serializes all I/O and adds more than one minute of startup delay. This delay has caused availability issue and I had to bump the timeout to mitigate. 

### Benchmark Results

50,000 files × 10 KiB each, ext4, cold page cache:

| Method | Time |
|---|---|
| Old behavior (io.Copy, no fsync) | 4.4s |
| New sparse-aware, **no fsync** | 4.8s |
| New sparse-aware, single sync at end | 4.9s |
| New sparse-aware, **per-file fsync (current)** | **1m 16s** |

The per-file fsync makes the copy **17x slower**. The sparse-aware logic itself (SEEK_DATA/SEEK_HOLE + copy_file_range) adds only ~10% overhead vs the old io.Copy path.

Full writeup and reproduction code: https://github.com/pengubco/ad-hoc-continuity-0.5.0-fsync-regression

### Impact

This affects any code path that uses `fs.CopyDir` or `fs.CopyFile` for directories with many files:

- **Volume copy-up** — containerd CRI plugin copies image content into volume mounts at container start (`volumeCopyUp`). Related: #8639 in containerd/containerd.
- **`nerdctl cp`** — copying files into/out of containers.
- **Snapshot operations** — any directory tree replication using `continuity/fs`.

### Why per-file fsync is unnecessary

The old `openAndCopyFile` (io.Copy) never called fsync. For the primary callers of `CopyDir`:

- Volume copy-up: if the machine crashes, the container never started — the runtime redoes the copy from the immutable image layer on restart.
- Image extraction: interrupted extractions are discarded and re-extracted from the content store.

None of these paths require per-file durability. The data source is always available for replay.

### Suggestion

Remove `tgt.Sync()` from `copyFile`. If durability before return is desired for some callers, a single `sync()` at the end of `CopyDir` provides equivalent crash safety with <1% overhead (vs 17x with per-file). Alternatively, make it opt-in via a `CopyDirOpt`.

### Environment

- Linux (ext4)
- Go 1.26
- `continuity` at commit a424ba1
- AWS EC2 with EBS volumes

ayush-panta:
Fixes #303

## Problem

The per-file `tgt.Sync()` added in a424ba1 causes significant latency for workloads that copy many files. Most callers of `CopyFile` and `CopyDir` do not require per-file durability, as the source data is typically available for replay.

## Changes
- Add `WithFileSync()` option to `CopyFile` for opt-in per-file fsync
- Add `WithCopyFileSync()` and `WithCopyDirSync()` options to `CopyDir`
- Default behavior is no-fsync, restoring behavior prior to a424ba1
- `WithCopyDirSync()` implies file-level fsync and additionally fsyncs each directory after all entries are copied

The variadic signature for `CopyFile` is backwards-compatible and existing callers require no changes.

## Testing
Added unit tests. Additionally tested by vendoring into containerd (main) and running the blockfile snapshotter, archive, and CRI podsandbox test suites successfully.

### DIFF
--- a/fs/copy.go
+++ b/fs/copy.go
@@ -33,7 +33,9 @@ type XAttrErrorHandler func(dst, src, xattrKey string, err error) error
 type copyDirOpts struct {
 	xeh XAttrErrorHandler
 	// xex contains a set of xattrs to exclude when copying
-	xex map[string]struct{}
+	xex      map[string]struct{}
+	fileSync bool
+	dirSync  bool
 }
 
 type CopyDirOpt func(*copyDirOpts) error
@@ -65,6 +67,34 @@ func WithXAttrExclude(keys ...string) CopyDirOpt {
 		for _, key := range keys {
 			o.xex[key] = struct{}{}
 		}
+		return nil
+	}
+}
+
+// WithCopyFileSync ensures each file copied within CopyDir is fsynced to
+// persistent storage before proceeding to the next file.
+//
+// By default, files are not synced. Versions prior to v0.5.0 never synced;
+// v0.5.0 added unconditional per-file fsync which caused a performance
+// regression.
+func WithCopyFileSync() CopyDirOpt {
+	return func(o *copyDirOpts) error {
+		o.fileSync = true
+		return nil
+	}
+}
+
+// WithCopyDirSync ensures full crash durability during CopyDir.
+// This implies file-level fsync (WithCopyFileSync) and additionally fsyncs
+// each directory after all its entries have been copied.
+//
+// By default, neither files nor directories are synced. Versions prior to
+// v0.5.0 never synced; v0.5.0 added unconditional per-file fsync which
+// caused a performance regression.
+func WithCopyDirSync() CopyDirOpt {
+	return func(o *copyDirOpts) error {
+		o.fileSync = true
+		o.dirSync = true
 		return nil
 	}
 }
@@ -143,7 +173,7 @@ func copyDirectory(dst, src string, inodes map[uint64]string, o *copyDirOpts) er
 				if err := os.Link(link, target); err != nil {
 					return fmt.Errorf("failed to create hard link: %w", err)
 				}
-			} else if err := CopyFile(target, source); err != nil {
+			} else if err := copyFileWithOpts(target, source, o); err != nil {
 				return fmt.Errorf("failed to copy files: %w", err)
 			}
 		case (fileInfo.Mode() & os.ModeSymlink) == os.ModeSymlink:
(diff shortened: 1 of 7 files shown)


fs/sync_windows.go
```diff
@@ -0,0 +1,24 @@
+//go:build windows
```
AkihiroSuda: unneeded line

fs/copy.go
```diff
@@ -185,16 +207,51 @@ func copyDirectory(dst, src string, inodes map[uint64]string, o *copyDirOpts) er
...
+}
+
+type CopyFileOpt func(*copyFileConfig)
+
+type copyFileConfig struct {
+	sync bool
+}
+
+// WithFileSync ensures the copied file is fsynced to persistent storage before returning.
+// By default, files are not synced to avoid unnecessary I/O latency.
```
AkihiroSuda: Could you document the default behavior in the past versions too

fs/copy.go
```diff
@@ -69,6 +71,26 @@ func WithXAttrExclude(keys ...string) CopyDirOpt {
...
+func WithCopyFileSync() CopyDirOpt {
+	return func(o *copyDirOpts) error {
+		o.fileSync = true
+		return nil
+	}
+}
+
+// WithCopyDirSync ensures full crash durability during CopyDir.
+// This implies file-level fsync (WithCopyFileSync) and additionally fsyncs
+// each directory after all its entries have been copied.
```
AkihiroSuda, line 89: Could you document the default behavior in the past versions too

AkihiroSuda (approved):

dcantah (approved):

AkihiroSuda (approved):

dmcgowan:
Other sync PR got merged first, can you rebase this one

dmcgowan:
Rebased it, I think this one makes more sense than the one that just removed it anyway.

ayush-panta:
Hi @dmcgowan, can we get this merged so that it gets picked up in the next release? I'm a bit concerned it might slip through the cracks as the associated issue has been closed.

AkihiroSuda (approved):

PR #320, opened 2026-08-21
feature: use bloom to reduce cpu usage

ningmingxiao:


AkihiroSuda:
Any benchmark result?

fs/du_unix.go
```diff
@@ -58,12 +61,46 @@ func fileInfoStat(path string, fi os.FileInfo) (*syscall.Stat_t, error) {
...
+	return arr[:]
+}
+
 func diskUsage(ctx context.Context, roots ...string) (Usage, error) {
 	var (
 		size   int64
 		inodes = map[inode]struct{}{} // expensive!
 	)
-
+	store := NewInodeStore(500000, 0.01)
```
AkihiroSuda, line 103: How were these parameters decided?

ningmingxiao: the  bottleneck is system calls   it's a estimated value see https://github.com/bits-and-blooms/bloom , after some test this pr is not helpful for https://github.com/containerd/containerd/issues/13967   @AkihiroSuda

